PR #21, opened 2020-12-15
Add support for raw JQL

hellosab:
User should be able to pass raw JQL, user-supplied JQL should override other filters if passed to prevent confusion 

example: 

```
/jira-cli issue -j "assignee in (currentuser())"
```

### DIFF
--- a/internal/query/issue.go
+++ b/internal/query/issue.go
@@ -73,6 +73,9 @@ func (i *Issue) Get() string {
 	if i.params.debug {
 		fmt.Printf("JQL: %s\n", q.String())
 	}
+	if i.params.jql != "" {
+		q.And(func() { q.Raw(i.params.jql) })
+	}
 
 	return q.String()
 }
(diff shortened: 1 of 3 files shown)


ankitpokhrel (commented):
@hellosab Thank you for the PR!

This is breaking `epic` and `sprint <ID>` cmd. As of now, `epic` and `sprint <ID>` supports all flags available in the `issue` cmd so we might need to make `jql` flag available in those cmd too.

See:
- [cmd/jira/epic.go](https://github.com/ankitpokhrel/jira-cli/blob/master/cmd/jira/epic.go#L120)
- [cmd/jira/sprint.go](https://github.com/ankitpokhrel/jira-cli/blob/master/cmd/jira/sprint.go#L130)

internal/query/issue.go
```diff
@@ -73,6 +73,9 @@ func (i *Issue) Get() string {
 	if i.params.debug {
 		fmt.Printf("JQL: %s\n", q.String())
 	}
+	if i.params.jql != "" {
```
ankitpokhrel, line 76: One blank line above will make the linter happy :)

ankitpokhrel:
Closing since the codebase has diverged a lot.

PR #101, opened 2021-05-22
Removed emoji, some color addition

qualious:
Hey! 

I know this is a wildcard PR but wanted to create a PR from the local changes that I've made. I tried using font glyphs but they didn't work so I've decided to use the pipe symbol instead. It's an opinionated thing, I don't like emoji on TUI. Maybe make it optional? 

I don't know if this is helpful at all, but wanted to add it anyways.


.gitignore
```diff
@@ -9,3 +9,5 @@
 dist/
 coverage/
 vendor/
+tags
+tags.*
```
samundra, line 13: I don't think these are required on project specific `.gitignore`. These should be mentioned in your global git ignore https://docs.github.com/en/github/getting-started-with-github/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

samundra:
I agree that PR looks very very opinionated. IMHO it doesn't add any feature or bring any meaningful additions to project. Changes like these will always be based on ones preferences and thus should be handled as such.

I disagree on adding these kind of changes to project, rest is upto @ankitpokhrel whether he wants to merge it or not.

qualious:
@samundra 

Hey! Thanks for that global gitignore mention.

-- 

I agree that PR adds nothing and is just a quick hack that I've done locally which is not ready to merge to the actual repo.

The reason I opened the PR is generally asking for options such as these. I can also implement them myself and I realize that's kinda the point but I'm already burned out from my day job, I honestly don't know. Maybe in the future I can implement these as options such as:

use_emojis = 0
use_alternating_rows = 1
selected_row_style = ...

etc. 

Please let me know if such configuration is desirable and please ignore the PR otherwise.

.gitignore
qualious: you are right! done it as such. thanks!

ankitpokhrel:
@qualious Thank you for your contribution! 

Allowing a user to control each part of the UI is not ideal and is out of the current scope of the project. Having said that, there is a feature in the pipeline to implement `--no-color` option that will disable colors + emojis. I am happy to review if someone wants to work on the feature.

I am closing this PR but feel free to [open a discussion](https://github.com/ankitpokhrel/jira-cli/discussions/new) in case you want to discuss this further. 

Thank you again!

PR #134, opened 2021-09-09
Update browser module to latest.

fraggerfox:
This prevents build errors in OSes where the module does not have support.

`../../pkg/mod/github.com/pkg/browser@v0.0.0-20210706143420-7d21f8c997e2/browser_unsupported.go:7:2: imported and not used: "os/exec"`

Related Issue: https://github.com/pkg/browser/issues/37

NOTE: Also a PR to support NetBSD in browser has been submitted https://github.com/pkg/browser/pull/39

PR #137, opened 2021-09-11
Update browser module - Adds NetBSD support.

fraggerfox:
Looks like https://github.com/pkg/browser/pull/39 got merged and now there should be native NetBSD support.

Updated the module to point to the latest.

Reference: Build fix was done in PR https://github.com/ankitpokhrel/jira-cli/pull/134


ankitpokhrel (approved):
Thank you!

PR #151, opened 2021-10-01
docs: Added how to use vim for editing description

ktrueda:
Hi !

I love this project so much.
My first question after installation is `How can I use vim for editing description?`.
I believe many users also think so.

I read [this code](https://github.com/ankitpokhrel/jira-cli/blob/c7d57ccb883f656403b5a3fa51b87077d5a0ce49/pkg/surveyext/editor.go#L24) and use VISUAL.

If readme shows how to do that, it may be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,9 @@ $ jira issue create
 
 # Pass required parameters to skip prompt or use --no-input option
 $ jira issue create -tBug -s"New Bug" -yHigh -lbug -lurgent -b"Bug description"
+
+# Create an issue using interactive prompt (using vim to edit description)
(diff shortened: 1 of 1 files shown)


ankitpokhrel:
Hi @ktrueda, Thank you for the PR. Most CLI apps will allow you to override the editor using `EDITOR` or `VISUAL` env. I don't think we need to mention it explicitly in the create command since this is also true for all commands that use JiraEditor eg: edit, comment, etc.

I moved this information to the FAQs section [here](https://github.com/ankitpokhrel/jira-cli/discussions/154).

ktrueda:
great!
thanks

Issue #143, opened 2021-09-14
Ability to list issues in several projects


ecksun:
I would like to list issues spanning multiple projects, for example corresponding to this JQL query:

    project in (APP, WEB, AI) AND labels in (backend)

Perhaps by providing a comma separated list to `--project`?

ankitpokhrel:
@ecksun Right now we are binding JQL to a single project context. I think it will be better to provide a way to execute any valid JQL. Since project is a required field what we can do is:
- Check if `project` key exist in provided JQL
- If yes, we will use the query as is
- If not, we will append currently configured project to the query

It should then allow you to use any query combination.

ecksun:
Ah, so that would mean that this would NOT be possible:

    jira issue list --label backend --project APP,WEB,AI

but

    jira issue list --jql 'project in (APP, WEB, AI) AND labels in (backend)'

would?

That works, not as convenient but JQL is not that bad to write

PR #153, opened 2021-10-03
:sparkles: Increased issue list project scope

ctreminiom:
1. Enabled the possibility to list issues from different projects.
2. Added a new flag called projects that increases the resulting scope.

In the current behavior, the issue is taken from the default project key we selected on the **init** command. It does not enable the user to merge the issue result using different projects.

### Actual Behavior
issues-without-filter

With this PR, the sub-command provides the flag **projects** enables the user to increase the list context searching and render issues from multiples projects.

`jira issue list --projects DESK --projects PV`

issues-with-filter

Also, if you want to omit the default project, you can replace it using the **project** flag

`jira issue list --projects DESK --project PV  --plain`


ctreminiom:
Related to #143 :rocket:

ankitpokhrel:
Hi @ctreminiom, Thank you for working on this PR.

Instead of introducing a new flag, I think the better solution would be to provide a way to execute any valid JQL irrespective of the project setup in the config. This way, anyone would be able to run raw JQL as per their need to list entries.

Feel free to open a [thread here](https://github.com/ankitpokhrel/jira-cli/discussions) in case you want to discuss the solution and proceed with the change.

ankitpokhrel:
Closing since the main branch has diverged a lot. Thank you again for your contribution.

PR #171, opened 2021-10-31
Fix example URL spelling in generator.go

DannyJJK:
Fix example URL spelling of Atlassian


ankitpokhrel (approved):
Nice catch 👀 

Thank you!

PR #238, opened 2021-12-31
fix #237: add targets: clean, distclean

### LINKED ISSUE #237: edit Makefile: add `clean` target
**Is your feature request related to a problem? Please describe.**
There is no `clean` target in the `Makefile`. 

**Describe the solution you'd like**
A `clean` target in the `Makefile` that supports the deletion of build artifacts from the git worktree. This may require overriding `GOCACHE` such that build artifacts are actually created within the git worktree.

**Describe alternatives you've considered**
* `go clean ./...`
* `git clean -ffxd`

**Additional context**
The `clean` target is one of the Standard Targets mentioned in the GNU Make Manual. Having such a target is A Good Thing.

References:
* https://www.gnu.org/software/make/manual/make.html#Standard-Targets

realtime-neil:
* edit .gitignore: add /.gocache

* edit Makefile:

  * add .PHONY targets clean and distclean

  * export GOCACHE with default

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ dist/
 bin/
 coverage/
 vendor/
+/.gocache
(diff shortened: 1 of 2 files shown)


ankitpokhrel (approved):
Thank you!

PR #249, opened 2022-01-06
build: Initialize version using runtime build info

ankitpokhrel:
This PR adds a file that is compiled only for go1.12 or later. The main purpose is to initialize `version` from runtime information when user runs `go install github.com/ankitpokhrel/jira-cli/cmd/jira@latest`. Version is populated based on the latest build version.

```sh
(Version="v0.2.1-0.20220106190958-d362a5d3dedc", GitCommit="", CommitDate="", GoVersion="go1.16.6", Compiler="gc", Platform="linux/amd64")
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,10 @@ brew install jira-cli
 ```
 
 #### Manual
-You can also install the runnable binary to your `$GOPATH/bin`.
+You can also install the runnable binary to your `$GOPATH/bin` (go1.16+).
 
 ```sh
-go install github.com/ankitpokhrel/jira-cli/cmd/jira
+go install github.com/ankitpokhrel/jira-cli/cmd/jira@latest
 ```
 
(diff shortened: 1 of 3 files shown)


realtime-neil:
Oh, I like that. I had no idea this kind of support existed for golang repositories.

realtime-neil:
@ankitpokhrel what's causing that discrepancy in the version patch number?

ankitpokhrel:
I've no clear idea TBH but I think it has something to do with version 0. The [document](https://go.dev/doc/modules/version-numbers#in-development) says.

> A module published with a v0 number will have a formal semantic version number with a major, minor, and patch part, as well as an optional pre-release identifier.

So I assume it should have been `v0.2.0-0.20220107075457-71969fade44c` 🤷🏻‍♂️

I will dig into this later.

PR #269, opened 2022-01-14
Feat/improve api tokens

jonathanverner:
This pull requests contains two improvements to API TOKEN handling:

1) The first improvement (commit 957284f) allows reading the api token from an external command
2) The second improvement allows using PAT tokens with an on-premise (local) jira server

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -17,15 +17,20 @@ const clientTimeout = 15 * time.Second
 
 var jiraClient *jira.Client
 
-// GetTokenConfig returns the API token config value amd whether it is a command.
-func GetTokenConfig() (string, bool) {
+// GetTokenConfig returns the API token config value, whether it is a command,
+// and whether it is a PAT.
+func GetTokenConfig() (string, bool, bool) {
 	switch {
 	case viper.GetString("api_token") != "":
-		return viper.GetString("api_token"), false
+		return viper.GetString("api_token"), false, false
 	case viper.GetString("api_token_cmd") != "":
-		return viper.GetString("api_token_cmd"), true
+		return viper.GetString("api_token_cmd"), true, false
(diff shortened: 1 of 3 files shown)


api/client.go
```diff
@@ -13,6 +17,23 @@ const clientTimeout = 15 * time.Second
...
+		return viper.GetString("api_token"), false, false
+	case viper.GetString("api_token_cmd") != "":
+		return viper.GetString("api_token_cmd"), true, false
+	case viper.GetString("api_token_pat") != "":
+		return viper.GetString("api_token_pat"), false, true
+	case viper.GetString("api_token_pat_cmd") != "":
+		return viper.GetString("api_token_pat_cmd"), true, true
+	default:
+		return "", false, false
+	}
```
ankitpokhrel, line 34: I was thinking if we can do this a bit differently.
- It might be better to use same `JIRA_API_TOKEN` env for both basic and bearer auth. The common usecase would be either to use basic or bearer.
- A new env `JIRA_AUTH_TYPE` which could specify whether it is a `bearer` or `basic` auth (basic by default).
- A new env `JIRA_AUTH_CMD` which will be invoked to get the token if it is set.

Also, I wouldn't add tokens directly in the config file. One of the idea behind config file for me was also to be able to share it with colleagues or commit it in vcs and storing tokens in the config defeats that purpose.

ankitpokhrel (commented):
Hi @jonathanverner, Thank you for working on this 🙏🏻

I added a few comments and I think the implementation can be simplified a bit. Let me know what you think.

ankitpokhrel:
Closing due to inactivity. The branch has diverged a bit and some part of this PR is done in #327. Please feel free to comment or open a new PR for further discussion.

PR #285, opened 2022-02-03
fix #277

### LINKED ISSUE #277: 404 error: jira issue list --limit=$((1<<31))
**Describe the bug**
The upper limit on the `jira issue list --limit` is an undocumented 2147483647; i.e., 2**31 - 1

**Please provide following details**
1. JiraCLI Version:
    ```
    $ jira version
    (Version="v0.3.1-0.20220124201418-5429cbd9d363", GitCommit="", CommitDate="", GoVersion="go1.17.6", Compiler="gc", Platform="linux/amd64")
    ```
2. Are you using Jira cloud or on-premise jira server? Also mention the version for on-premise installation.
    ```
    cloud
    ```
3. What operating system are you using? Also mention version.
    ```
    Ubuntu 20.04
    ```
4. What terminal are you using? Also mention version.
    ```
    $ gnome-terminal --version
    # GNOME Terminal 3.36.2 using VTE 0.60.3 +BIDI +GNUTLS +ICU +SYSTEMD
    ```

**To Reproduce**

```
$ jira issue list --plain --limit="$(((1 << 31) - 1))" >/dev/null 2>&1 ; echo $?
0
$ jira issue list --plain --limit="$(((1 << 31) - 0))" ; echo $?
⠹ Fetching issues... 
jira: Received unexpected response '404 Not Found'.
Please check the parameters you supplied and try again.
1
```

**Expected behavior**

Some kind of error message blaming the size of the limit, and not an error message reporting a 404.

**Screenshots**

Not applicable.

**Additional context**

Not applicable.

realtime-neil:
edit pkg/jira/search.go: clamp limit to MaxInt32

### DIFF
--- a/pkg/jira/search.go
+++ b/pkg/jira/search.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"net/url"
 )
@@ -32,6 +33,11 @@ func (c *Client) search(jql string, limit uint, ver string) (*SearchResult, erro
 		err error
 	)
 
+	// maxResults is a int32
+	// https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get
+	if math.MaxInt32 < limit {
+		limit = math.MaxInt32
+	}
 	path := fmt.Sprintf("/search?jql=%s&maxResults=%d", url.QueryEscape(jql), limit)
 
 	switch ver {


pkg/jira/search.go
```diff
@@ -32,6 +33,11 @@ func (c *Client) search(jql string, limit uint, ver string) (*SearchResult, erro
 		err error
 	)
 
+	// maxResults is a int32
+	// https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get
+	if math.MaxInt32 < limit {
+		limit = math.MaxInt32
+	}
```
ankitpokhrel, line 40: I am not sure if this fix is necessary given that Jira API won't return the expected number of records even for numbers like 120 or 150 as the response payload is huge. The output we get when using the `limit` option is not reliable, and I think this feature needs to be redesigned to accept a range of items (eg: from 100, take 50).

realtime-neil: @ankitpokhrel point taken. Is there an issue for the problem where the returned payload < limit ? What recourse do we have if the Jira endpoint won't return _all_ the list items we request?

ankitpokhrel: @realtime-neil As per [the docs](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get)

```
The greatest number of items returned per page is achieved when requesting id or key only.
```

But if we only request for `id` or `key` to make the limit functionality work, we would need to send separate requests to get other details we are displaying in the TUI. And that's not feasible.

realtime-neil: point taken

realtime-neil:
@ankitpokhrel leaving aside the problem with the Jira API endpoints returning fewer-than-expected items, how do you want to handle the inherent discrepancy between the types given to the `jira-cli` option `--limit` and the Jira API option `maxResults` (`uint32` and `int32`, respectively)?

ankitpokhrel:
@realtime-neil I haven't given enough thought to this but I would probably remove `limit` option and introduce something like `from` and `take` with a cap of perhaps 100 items (depends on behavior of the API). This will also allow us to implement pagination functionality in the TUI.

ankitpokhrel:
`limit` flag no longer exists so this PR is not relevant anymore

PR #297, opened 2022-02-11
Allow user to set default priority, labels & components

frankpf:
Read `default_components`, `default_labels` and `default_priority` from the `project` key in config and use those as the default when creating an issue.


ankitpokhrel:
Hi @frankpf, thank you for your work. I don't see a good reason to have this change in the tool since this can be achieved by creating an alias of some sort. Please feel free to comment on the ticket if you think otherwise.

Thank you again for your contribution!

PR #300, opened 2022-02-16
feature: Append label to an issue, show labels at issue list view. 

stchar:
Sometimes I just need to add a new label to an issue, unfortunately `jira issue edit` is designed to overwrite them.

Another miss is if I want to get the labels I need to open the browser or call `jira issue view`
```
jira issue edit -lLABUG --label-append
```


ankitpokhrel:
Hi @stchar, Thank you for the PR.

I can agree that the default behavior for label edit should be to append the labels so that we won't accidentally lose the data. But, Instead of introducing a new flag `label-append`, I would go with changing the current behavior of the flag `label` to append the labels instead of overriding them (better than accidentally overriding the data IMO). 

Also, let's keep table view as is. We can do a wide variety of things with Jira, and it may not be possible to incorporate everything, and I would like to keep things as simple as possible. A better thing to do would be to make the tool extensible so that devs can add new features on top of the existing one (using plugins).

stchar:
Hi @ankitpokhrel, thanks for the input. I've removed the labels field form the issue view. 

Regarding `label-append` I have a concern that if I change default behavior of -l flag which is now rewrites the label filed. I'll break the backward compatibility.  

And people who  already use it a lot they may be frustrated.  Because they are already ok that it rewrites the field and may have some scripts on top of it. So now its not an architecture miss rather than a feature )) That's why I added additional flag to change the behaviour. What do you think?

ankitpokhrel:
@stchar that's a valid concern, but we are still in v0, and as per semantic versioning, that's acceptable. Be assured that we won't have any breaking change once we reach v1.

From [the doc](https://semver.org/#spec-item-4)
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

stchar:
@ankitpokhrel, ok cool, will rework it.

stchar:
@ankitpokhrel, it's done.

ankitpokhrel (commented):
@stchar Just one small comment, looks good otherwise. Thank you!

stchar:
@ankitpokhrel, cannot see your comments. Did you submit the review?

internal/cmd/issue/edit/edit.go
```diff
@@ -341,7 +343,7 @@ func setFlags(cmd *cobra.Command) {
 	cmd.Flags().StringP("body", "b", "", "Edit description")
 	cmd.Flags().StringP("priority", "y", "", "Edit priority")
 	cmd.Flags().StringP("assignee", "a", "", "Edit assignee (email or display name)")
-	cmd.Flags().StringArrayP("label", "l", []string{}, "Replace labels")
+	cmd.Flags().StringArrayP("label", "l", []string{}, "Add labels to existed list")
```
ankitpokhrel, line 346: Sorry, forgot to submit.

```suggestion
	cmd.Flags().StringArrayP("label", "l", []string{}, "Append labels")
```

ankitpokhrel (approved):

PR #301, opened 2022-02-16
feat(make) install binaries to ./bin

stchar:
I had no experience with go before. So for me it was not so obvious where can I get `jira` executable after I run `make deps install`


ankitpokhrel:
`GOBIN` by default should be `$GOPATH/bin`.

PR #305, opened 2022-02-20
feat: Allow insecure TLS

ankitpokhrel:
This PR adds an `--insecure` option in `init` cmd to skip TLS verification. This can be useful if the Jira server is using self-signed certificates.

Relates to: https://github.com/ankitpokhrel/jira-cli/discussions/271#discussioncomment-2091603

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -28,8 +28,13 @@ func Client(config jira.Config) *jira.Client {
 	if config.APIToken == "" {
 		config.APIToken = viper.GetString("api_token")
 	}
+	config.Insecure = viper.GetBool("insecure")
 
-	jiraClient = jira.NewClient(config, jira.WithTimeout(clientTimeout))
+	jiraClient = jira.NewClient(
+		config,
+		jira.WithTimeout(clientTimeout),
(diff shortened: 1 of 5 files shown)


MattPD:
I've just checked this option after the 1.0.0 release, but I cannot go past the `init` step:

```
jira init --insecure
You are using --insecure option. In this mode, the client will NOT verify
server's certificate chain and host name in requests to the jira server.

? Config already exist. Do you want to overwrite? Yes
? Installation type: Local
? Link to Jira server: (MY_SERVER)
? Login username: (MY_USERNAME)

✗ Unable to generate configuration: x509: certificate signed by unknown authority
```

FWIW , the previous workaround (in `github.com/ankitpokhrel/jira-cli/cmd/jira@patch-tls-issue`) worked for me.

ankitpokhrel:
Hi @MattPD, it looks like an issue with config generator. Does it work for you if you manually add this line in your config yaml file?

```
insecure: true
```

MattPD:
Thanks, yes, adding `insecure: true` to the `~/.config/.jira/.config.yml` did the trick!

Issue #312, opened 2022-02-23
Implement bearer authentication with personal access tokens


marek-veber:
The current implementation uses basic authentication only.
But our server enables only bearer authentication :(
See https://confluence.atlassian.com/enterprise/using-personal-access-tokens-1026032365.html

**Describe the solution you'd like**
Implement the `bearer` flag into the config file. The juice if the `JIRA_AUTH_TOKEN` contains a `password` or a `personal access tokens` will be answered during the init process. 
The code for auth will be changed to:
```
	if c.bearer {
		req.Header.Add("Authorization", "Bearer "+c.token)
	} else {
		req.SetBasicAuth(c.login, c.token)
	}
```

**Describe alternatives you've considered**
The suggested auth code is simple. The solution how to create the `bearer` flag may differ from the description above.

**Additional context**

marek-veber:
@ankitpokhrel - many thanks for finishing this issue.

PR #313, opened 2022-02-23
Add the bearer authentication

marek-veber:
Closes [312](https://github.com/ankitpokhrel/jira-cli/issues/312)

### DIFF
--- a/pkg/jira/client.go
+++ b/pkg/jira/client.go
@@ -96,6 +96,7 @@ type Config struct {
 	Server   string
 	Login    string
 	APIToken string
+	Bearer   *bool
 	Insecure bool
 	Debug    bool
 }
@@ -107,6 +108,7 @@ type Client struct {
 	server    string
 	login     string
 	token     string
+	bearer    bool
 	timeout   time.Duration
 	debug     bool
 }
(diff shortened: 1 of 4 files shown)


pkg/jira/client.go
```diff
@@ -96,6 +96,7 @@ type Config struct {
 	Server   string
 	Login    string
 	APIToken string
+	Bearer   *bool
```
ankitpokhrel, line 99: Enum types would be better for this usecase instead of a boolean flag. This will make it easy to introduce other auth types if needed.

```go
type AuthType string

const (
	AuthTypeBasic  AuthType = "basic"
	AuthTypeBearer AuthType = "bearer"
)
```

And the `config` and `client` struct downstream can have `AuthType` field

ankitpokhrel:
Hi @marek-veber, thank you for the PR. I added some comments.

Instead of asking for auth type from the UI, I think It might be better to introduce a new env `JIRA_AUTH_TYPE` which could specify what authentication method to use: `bearer` or `basic` (basic by default). Majority of the users seem to use basic auth and we could reduce one additional steps for them if we go with environment variable approach.

Also, we need to make sure this works for both cloud and on-premises jira server.

marek-veber:
Thanks for your comments, I will implement the introduced changes.

Issue #292, opened 2022-02-08
Feature: Issue List Refresh Interval


jbaskeen:
**Feature Request**

After reviewing the options I could not find a way to have an issue list, manually or automatically, refresh itself. After opening an issue list in my terminal, I'd like to be able to come back to it later and know that it is current, without having to back out and run the command again.

Provide a way to:

1. Manually refresh a list while viewing
2. Provide a refresh interval flag which accepts a counter which triggers a view refresh

If this feature does exists, let's document it or make it more obvious.

ankitpokhrel:
Hi @jbaskeen, this could be an interesting feature to add. I do miss this feature myself sometimes. Things are bit overwhelming for me right now but I might work on this in the near future. PRs are welcome in the mean time :)

ankitpokhrel:
Hi @jbaskeen, point 1 is done with #325. You can use `CTRL+R` or `F5` to refresh the list. This will be available in the next release. If you wish to try it now, you can either compile the [main branch manually](https://github.com/ankitpokhrel/jira-cli#development) or if you are using [Homebrew](https://brew.sh/) you can use `brew install jira-cli --HEAD`.

PR #325, opened 2022-03-10
feat: support refresh for issues list

GZLiew:
This PR add refreshing behaviour when the `<C-R>` or `F5` key is pressed when viewing issues list

relating to #292 but with no interval implemented

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ The lists are displayed in an interactive UI by default.
 - Use arrow keys or `j, k, h, l` characters to navigate through the list.
 - Use `g` and `SHIFT+G` to quickly navigate to the top and bottom respectively.
 - Press `v` to view selected issue details.
-- Press `CTRL+R` to refresh issues list.
+- Press `CTRL+R` or `F5` to refresh issues list.
 - Hit `ENTER` to open the selected issue in the browser.
 - Press `c` to copy issue URL to the system clipboard. This requires `xclip` / `xsel` in linux.
 - Press `CTRL+K` to copy issue key to the system clipboard.
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ The lists are displayed in an interactive UI by default.
 - Use arrow keys or `j, k, h, l` characters to navigate through the list.
 - Use `g` and `SHIFT+G` to quickly navigate to the top and bottom respectively.
 - Press `v` to view selected issue details.
+- Press `r` to refresh issues list.
 - Hit `ENTER` to open the selected issue in the browser.
 - Press `c` to copy issue URL to the system clipboard. This requires `xclip` / `xsel` in linux.
 - Press `CTRL+K` to copy issue key to the system clipboard.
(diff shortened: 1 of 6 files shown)


pkg/tui/table.go
```diff
@@ -165,6 +176,13 @@ func (t *Table) initTable() {
 			}
 		}).
 		SetInputCapture(func(ev *tcell.EventKey) *tcell.EventKey {
+			if ev.Rune() == 'r' {
```
ankitpokhrel, line 179: Let's use standard `CTRL + R` / `F5` hotkeys for this

```suggestion
			if ev.Key() == tcell.KeyCtrlR || ev.Key() == tcell.KeyF5 {
```

ankitpokhrel (commented):
Hi @GZLiew, Thank you for working on this 🙌🏻

I wanted to do this differently but that would require some rework on how the TUI is implemented. So, I think your approach is fine for now and I will improve it later.

I do have some comments tho.
1. Let's use standard `CTRL + R` / `F5` as a shortcut for refreshing.
2. Could you please extend this to `jira epic list <epic_id>` and `jira sprint list <sprint_id>`?

Thank you!

pkg/tui/table.go
GZLiew: Thanks for reviewing @ankitpokhrel ! alright I will proceed make a few more commits to address your comments above 👍🏻

README.md
```diff
@@ -138,6 +138,7 @@ The lists are displayed in an interactive UI by default.
 - Use arrow keys or `j, k, h, l` characters to navigate through the list.
 - Use `g` and `SHIFT+G` to quickly navigate to the top and bottom respectively.
 - Press `v` to view selected issue details.
+- Press `r` to refresh issues list.
```
adolsalamanca, line 141: Nitpick, this is still pointing to previous refresh shortcut.

pkg/tui/table.go
```diff
@@ -21,6 +21,9 @@ type SelectedFunc func(row, column int, data interface{})
 // ViewModeFunc sets view mode handler func which gets triggered when a user press 'v'.
 type ViewModeFunc func(row, col int, data interface{}) (func() interface{}, func(data interface{}) (string, error))
 
+// RefreshFunc is fired when a user press 'r' character in the table.
```
adolsalamanca, line 24: this comment should also be updated

pkg/tui/table.go
```diff
@@ -115,6 +119,13 @@ func WithViewModeFunc(fn ViewModeFunc) TableOption {
 	}
 }
 
+// WithRefreshFunc sets a func that is triggered when a user press 'r'.
```
adolsalamanca, line 122: Another one that requires update

adolsalamanca (commented):
Thanks for the work 👍🏽 (just another random contributor here) 
Could you please update Readme and comments so they stay up to date with latest shortcuts used?

README.md
GZLiew: yes of course! Sorry I missed that!

pkg/tui/table.go
GZLiew: thanks for pointing it out for me @adolsalamanca

ankitpokhrel (approved):
Thank you for working on this @GZLiew and thank you for reviewing @adolsalamanca, it helps! 🎉

Issue #188, opened 2021-11-08
Pick up the password/passkey from the .netrc file


cazador481:
Instead of having to put the passkey into an ENV variable, and store it in my dot files, can we pick also pick it up from the .netrc file?   I have a number of other tools that uses the .netrc file to store the jira keys and it would be a plus to have everything use the same location.

ankitpokhrel:
Hi @cazador481, I don't use `.netrc` so the chances of me implementing this anytime soon are quite low. However, if someone can send a PR (and maintain this part of the feature), I would be happy to review and get it merged.

ankitpokhrel:
This has been done in #329 and will be available in the next release. For now, you can either compile the [main branch manually](https://github.com/ankitpokhrel/jira-cli#development) or if you are using [Homebrew](https://brew.sh/) you can use `brew install jira-cli --HEAD`.

PR #329, opened 2022-03-13
Feature: add support to read from netrc

adolsalamanca:
## Description
This PR includes a tiny new feature to support reading configs from .netrc file as suggested in this [issue](https://github.com/ankitpokhrel/jira-cli/issues/188)

## Changes
* Create a .envrc reader under /pkg, credits to the folks from this [link](https://go.dev/src/cmd/go/internal/auth/netrc.go)
* Add support to jira-cli so it verifies .envrc config and also make jira calls rely on it in case it's valid, as a fallback to the token in `JIRA_API_TOKEN` env variable.
* Update checkForJiraToken function signature in order to check both possibilities now.
* Update README with new alternative use.

## Checks
- [x] Envrc reader has unit tests.
- [x] New feature is tested and working.

Let me know or feel free to edit the code if there's something that doesn't fit in terms of structure or approach :)

### DIFF
--- a/pkg/netrc/reader.go
+++ b/pkg/netrc/reader.go
@@ -7,7 +7,40 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 )
+
+var (
+	netrcOnce sync.Once
+	netrc     []netrcLine
+	netrcErr  error
+)
+
+// ReadNetrcPassword retrieves password for the desired server and login.
+// We just borrowed original code due to its internal nature, it is original implementation of .netrc read in Go
+// https://github.com/golang/go/blob/master/src/cmd/go/internal/auth/netrc.go
+func ReadNetrcPassword(serverURL string, login string) (string, error) {
+	URL, err := url.ParseRequestURI(serverURL)
+	if err != nil {
+		return "", err
+	}
+
+	netrcOnce.Do(func() {
+		readNetrc()
+	})
+
+	if netrcErr != nil {
+		return "", err
+	}
+
+	for _, line := range netrc {
+		if line.machine == URL.Host && line.login == login {
+			return line.password, nil
+		}
+	}
+
+	return "", errors.New("netrc token not found")
+}
 
 type netrcLine struct {
 	machine  string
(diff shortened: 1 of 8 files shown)


ankitpokhrel:
Hi @adolsalamanca, Thank you for working on this 🙌🏻 

Since the tool is getting some attention recently and lot of devs relying on the project, I think it would be better to stick with the [standard `.netrc` protocol ](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html)instead of reinventing the wheel. Lot's of CLI tool supports `.netrc` and I am not sure why we would need `envrc` instead of relying on `.netrc` file. Maybe there is a usecase that I am missing, let me know if that's the case.

As you mentioned in the description, golang already implements [netrc protocol](https://github.com/golang/go/blob/master/src/cmd/go/internal/auth/netrc.go). We can borrow the package (with credit) and use it for our token validation. We will have one less thing to maintain if we stick with this.

adolsalamanca:
Omg @ankitpokhrel, my bad, this was a typo.
It's actually implementing .netrc standard, just mixed up things with package name, branch name and also PR (will fix it now).

adolsalamanca:
Just borrowed the original implementation as it was from Go internals.
Only thing added was the function `ReadNetrcPassword` inside, not sure if could be better to extract to another package/file so we remain unchanged netrc files read implementation.

The only problem is that Deepsource is now complaining due to case "default" redundant inside netrc reader.

pkg/netrc/reader.go
```diff
@@ -0,0 +1,134 @@
...
+// https://github.com/golang/go/blob/master/src/cmd/go/internal/auth/netrc.go
+func ReadNetrcPassword(jiraServer string, login string) (string, error) {
+	netrcOnce.Do(func() {
+		readNetrc()
+	})
+	if netrcErr != nil {
+		return "", netrcErr
+	}
+
+	jiraServerURL, err := url.ParseRequestURI(jiraServer)
```
ankitpokhrel, line 30: netrc has nothing to do with Jira and `netrc` package should not care anything about it at all.

I would suggest to keep `reader` and `reader_test` sync with the [original implementation](https://github.com/golang/go/blob/master/src/cmd/go/internal/auth/netrc.go) including license information (required as per BSD).

You can have a new file called `netrc.go` in the package that will expose required methods to work with `netrc` package.

```go
// Package netrc implements GNU .netrc specification.
// This implementation is borrowed from the original implementation by the go authors.
// See https://github.com/golang/go/blob/master/src/cmd/go/internal/auth/netrc.go
// See https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
package netrc

import (
	"fmt"
)

// ErrNetRCEntryNotFound is thrown if details for the machine is not found.
var ErrNetRCEntryNotFound = fmt.Errorf("netrc: entry not found in config")

// Entry is a netrc config entry.
type Entry struct {
	Machine  string
	Login    string
	Password string
}

// Read reads config for the given machine.
func Read(machine string) (*Entry, error) {
	netrcOnce.Do(func() {
		readNetrc()
	})
	if netrcErr != nil {
		return nil, fmt.Errorf("netrc: %w", netrcErr)
	}

	for _, line := range netrc {
		if line.machine == machine {
			return &Entry{
				Machine:  line.machine,
				Login:    line.login,
				Password: line.password,
			}, nil
		}
	}

	return nil, ErrNetRCEntryNotFound
}
```

Now, you can use the exposed method elsewhere.

```go
netrcEntry, err := netrc.Read(config.Server)
```

Please feel free to update above implementation as necessary.

ankitpokhrel (commented):
Hi @adolsalamanca,

Overall this is getting in shape. There are few things we should keep in mind tho.

1. I believe `machine` name is unique in each `.netrc` file. If there are duplicates first entry is the valid one. These details are usually used for auto-login process and as per the spec, auto-login can be initiated as soon as the first entry for the machine is found.
2. When the user enters server name during `jira init` process, we need to make sure not to  ask for the `username/email` if it is present in `.netrc` file. We can do this in a different PR if this one is getting big.

Let me know if you need any help and thank you again for working on this.

ankitpokhrel:
> The only problem is that Deepsource is now complaining due to case "default" redundant inside netrc reader.
Let me know what you think about it.

Regarding this, don't worry about deepsource complaint in third-party or borrowed files. Not all issues reported by deepsource are crucial.

adolsalamanca:
Thanks for the reviews mate!

> netrc has nothing to do with Jira and netrc package should not care anything about it at all.

This is that happens when you refactor in automatic mode, but the original intention was completely separate responsibilities, that's why I created netrc in pkg folder. Good catch!

Will apply suggestions soon, have a good start of the week :)

adolsalamanca:
Hey @ankitpokhrel 
Thanks for your suggestions, I applied most of them.
However, I decided to keep login as a parameter for the reader, I know the chances to have two accounts for the same machine are pretty reduced, but this way it is mandatory to have the .netrc properly configured.

Still, there's something I would like to improve, when the `JIRA_API_TOKEN` is missing, and the .netrc config is not valid, the 403 could be a bit ambiguous, but we can iterate to improve that.

Thanks!!

ankitpokhrel (approved):
Hi @adolsalamanca, this seem to be working fine so merging this. Thank you 🙇🏻. 

Just a heads up, I might do some changes for better user experience.

PR #332, opened 2022-03-19
feature: Remote link command

emilekm:
Seeing that this command is missing (and it's mainly what I wanted to automate in my Jira workflow) I went on and implemented it.

This MR implements a bare minimum for creating a remote link with url and a title being required by the [API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-remote-links/#api-rest-api-3-issue-issueidorkey-remotelink-post).

I'm open for changing the way this works and adding more functionality if it's needed.
Thank you for this project.


emilekm:
I'm not sure how to fix the `dupl` linter check, I guess I could make a common function for testing these two that it is complaining.

PR #343, opened 2022-04-11
Fix authType vs AuthType var discrepancy

xbezdick:


ankitpokhrel:
Hi @xbezdick, Thank you for the PR!

I see that you changed unexported variables to the exported one. Could you please explain the reason for that? I don't see any discrepancies here.

xbezdick:
Hi, the unexported variable in go in for example Fedora is not available to client.go. It works on mac for some reason but will not work on most linux machines I tested on.

dofinn:
Confirmed this - `main` doesnt work on F36, but this PR does. Otherwise it defaults to "Basic" auth, regardless of setting env.

ankitpokhrel:
This is strange!

@xbezdick @danobi Can we confirm that `JIRA_AUTH_TYPE` variable is accessible in the shell/sub-shell and has a correct value? It looks like the variable is somehow not available in the shell. Do you have same issue if you set auth type directly in the config: `auth_type: bearer`?

xbezdick:
I think this was go-gcc vs golang. I'm no longer having this issue so I suggest closing this.

Issue #247, opened 2022-01-06
Support password manager for API token


d-miketa:
I'd rather not store my Jira API token in plain text. Would it be possible to add support for standard keyrings and password managers? Many thanks!

bspinner:
+1, I'd like to see interaction with the macOS keychain.

Storing tokens in plaintext in my env is a no-go for me :(

@ankitpokthrel What would be the TODOs for this?

ankitpokhrel:
@bspinner Keychain support is something in my TODO list. I can't give a date, but this will hopefully be done soon. There are other priorities at the moment.

If someone wants to contribute, we can utilize [this lib](https://github.com/zalando/go-keyring) to interact with keychain.

ankitpokhrel:
This is done with #348 and will be available in the next release. You can add a password item in your keychain with item name as `jira-cli` and account name as your login email/username. The order in which the tool look for a token is `env -> .netrc -> keychain`.

Please feel free to comment or open a new issue if this doesn't work for you.

PR #348, opened 2022-04-21
Add support for OS keyrings/-chains

boyvanamstel:
Addresses #247 

_Notice: Only tested on macOS._

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/ankitpokhrel/jira-cli/pkg/jira"
 	"github.com/ankitpokhrel/jira-cli/pkg/jira/filter"
+
+	"github.com/zalando/go-keyring"
 )
 
 const clientTimeout = 15 * time.Second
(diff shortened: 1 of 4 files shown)


api/client.go
```diff
@@ -36,6 +38,10 @@ func Client(config jira.Config) *jira.Client {
 			config.APIToken = netrcConfig.Password
 		}
 	}
+	if config.APIToken == "" {
+		secret, _ := keyring.Get("jira-cli", "token")
```
ankitpokhrel, line 42: Instead of `token` string, it might be better to use login email/username as a key like we do it with `.netrc`. This way you can configure different token for different user.

```suggestion
		secret, _ := keyring.Get("jira-cli", config.Login)
```

internal/cmd/root/root.go
```diff
@@ -157,6 +159,11 @@ func checkForJiraToken(server string, login string) {
 		return
 	}
 
+	secret, _ := keyring.Get("jira-cli", "token")
```
ankitpokhrel, line 162:
```suggestion
	secret, _ := keyring.Get("jira-cli", login)
```

ankitpokhrel (commented):
Hi @boyvanamstel, Thank you for working on this feature. I don't have a _windows_ machine to test this either but it seem to be working fine on macOS.

I added one small comment, looks good otherwise. Also, please don't forget to run `go mod tidy`.

boyvanamstel:
Thanks for the suggestions @ankitpokhrel. Have a look at the new commits. :)

ankitpokhrel (approved):
Thank you 🎉

PR #372, opened 2022-05-25
Support creating issues with custom subtask type

danobi:
Before, the code was assuming only one sub-task type existed (the
default). This would cause any issues filed with a custom sub-task type
to fail with the following error:

```
Error:
  - Issue 'XXX-23668' must be of type 'Epic'.

jira: Received unexpected response '400 Bad Request'
```

because `jira.CreateRequest::SubtaskField` was not being set.

Fix by modifying `GetSubtaskHandle()` helper to look for exact matches
on sub-task types.

### DIFF
--- a/internal/cmd/issue/create/create.go
+++ b/internal/cmd/issue/create/create.go
@@ -146,8 +146,8 @@ func create(cmd *cobra.Command, _ []string) {
 		}
 		cr.ForProjectType(projectType)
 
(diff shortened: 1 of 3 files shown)


ankitpokhrel (approved):
Ran into same issue today and this seems to be working fine. Thank you for working on this @danobi 🎉

PR #386, opened 2022-06-15
Only add project filter in JQL if project is not '' (empty string)

peterwwillis:
This change enables you to use `jira issue ..` commands - particularly, arbitrary JQL commands - without a project filter applied. Just pass an empty string for the project. *Probably* also works if you edit the config file to use an empty project name, but I haven't tried it.

For example:
```
$ jira issue list -p '' -q 'assignee IN (currentUser())' --plain
TYPE	KEY		SUMMARY									STATUS
Task	PROJ-123	One project task here							Backlog
Task	FOO-456		Different project here							Backlog

```

Above works for me to show me my current tickets across projects.

### DIFF
--- a/pkg/jql/jql.go
+++ b/pkg/jql/jql.go
@@ -24,7 +24,7 @@ type JQL struct {
 
 // NewJQL initializes jql query builder.
 func NewJQL(project string) *JQL {
-	var filterslist = []string{fmt.Sprintf("project=\"%s\"", project)}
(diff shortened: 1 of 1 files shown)


peterwwillis:
@ankitpokhrel thoughts?

ankitpokhrel:
@peterwwillis Thank you for the PR. While this works the solution doesn't seem that flexible. Eg: Doesn't allow us to filter by few specific projects.

Closing in favor of #395

PR #433, opened 2022-08-12
Add --summary flag to just print name of issue

blfordham:
Not sure if this is of interest, but I needed it and figured others might, too.

It adds a flag that causes the client to only print the summary of the issue.

eg: 
```$ jira issue view FOO-25 --summary
Do cool stuff```

### DIFF
--- a/internal/cmd/issue/view/view.go
+++ b/internal/cmd/issue/view/view.go
@@ -36,6 +36,7 @@ func NewCmdView() *cobra.Command {
 
 	cmd.Flags().Uint("comments", 1, "Show N comments")
 	cmd.Flags().Bool("plain", false, "Display output in plain mode")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by ankitpokhrel** [comment]
Hi @blfordham, you can get value for the issue summary using the following trick. I think that will work for your usecase.

```sh
jira issue list -q"key=ISS-1" --plain --no-headers --columns=summary | awk '{$1="";print $0}'
```

blfordham:
> ```shell
> jira issue list -q"key=ISS-1" --plain --no-headers --columns=summary | awk '{$1="";print $0}'
> ```

Yes, that would work. I had also done some bash magic, but I wasn't aware of some of the additional flags there. It also adds the issue key to the output, which I don't want, but that's pretty minor. Either way, since there are at least two ways to get this without the code I wrote, I'll close the PR.

Thanks for your help!

PR #467, opened 2022-09-12
Readme: add path for config file

snpefk:


ankitpokhrel:
Hi @snpefk, thank you for the PR. The config path may not always be the same. Running `jira -h` will list the correct config path for your setup. I think adding `$XDG_CONFIG_HOME/.jira/.config.yml` could be a bit confusing.

snpefk:
Thanks for reply, didn't see this. Maybe then I should add that `jira -h` message has info about current configuration? I was transferring my config between different machines, didn't find anything helpful in README.

PR #475, opened 2022-09-26
fix: Option to show issues from all projects in sprint list

### LINKED ISSUE #429: jira sprint list only shows issues from one board
jira version 1.0.0

For stories/bugs that are in the sprint but are set as Open status, `jira sprint list` or `jira sprint list --current` doesn't list those Open/Closed/PENDING/IMPLEMENTED/etc... Status, only listing only In Progress.

This shouldn't be the case, if we are performing `jira spring list` it should list all spring regardless of its' statuses and if they need to custom search the status a command such as `jira sprint list -sOpen` or `jira sprint list --status Closed` should be supported instead.

ankitpokhrel:
Fixes #429

This PR updates `jira sprint list` to list issues from all projects if `--show-all-issues` flag is provided. By default, the command only shows issues for the given/configured project.

```sh
$ jira sprint list --show-all-issues

$ jira sprint list <sprint-id> --show-all-issues
```

### DIFF
--- a/internal/cmd/sprint/list/list.go
+++ b/internal/cmd/sprint/list/list.go
@@ -78,17 +78,20 @@ func sprintList(cmd *cobra.Command, args []string) {
 
 	client := api.Client(jira.Config{Debug: debug})
 
+	sprintQuery, err := query.NewSprint(cmd.Flags())
+	cmdutil.ExitIfError(err)
+
 	if len(args) == 0 {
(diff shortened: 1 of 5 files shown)


ankitpokhrel:
Hi @pbrkr I don't have a fixed timeline but I will try to make a new release in few weeks.

pbrkr:
@ankitpokhrel Thank you!

PR #505, opened 2022-11-25
feat: filter issues by the absence of label(s)

martinpovolny:
you can now do e.g.

```
jira issues list -l~cna -l~web-rca -l~status-board -l~app-interface
```

to list issues that have none of the labels above

### DIFF
--- a/internal/query/issue.go
+++ b/internal/query/issue.go
@@ -32,6 +32,19 @@ func NewIssue(project string, flags FlagParser) (*Issue, error) {
 	}, nil
 }
 
+func splitPositiveNegative(labels []string) ([]string, []string) {
+	positive := make([]string, 0)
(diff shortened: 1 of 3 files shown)


ankitpokhrel (commented):
Hi @martinpovolny, this looks good. Could you please add a test to make sure `NotIn` method constructs a valid jql query?

martinpovolny:
Added the test, thx for the quick review!

ankitpokhrel (approved):
This is working great, thank you 🎉 

The mix of +ve and -ve labels is working good so we can do something like the following:

```sh
// Issues with `p1` and `q1` labels that doesn't have `urgent` label
jira issue list -lp1 -lq1 -l~urgent
```

PR #509, opened 2022-11-27
feat: allow setting of fixed columns in the list of issues, epics and sprints

martinpovolny:
Allow setting the number of locked (not scrolling) columns in the issue list, epic list and sprint list.

You can do e.g.:
```
jira issues list -s~Closed --fixed-columns 2
```

And the "Type" and "KEY" columns will not disappear as you press the right arrow.

### DIFF
--- a/internal/cmd/epic/list/list.go
+++ b/internal/cmd/epic/list/list.go
@@ -139,6 +139,9 @@ func singleEpicView(flags query.FlagParser, key, project, projectType, server st
 	noTruncate, err := flags.GetBool("no-truncate")
 	cmdutil.ExitIfError(err)
 
+	fixedColumns, err := flags.GetUint("fixed-columns")
+	cmdutil.ExitIfError(err)
+
 	columns, err := flags.GetString("columns")
 	cmdutil.ExitIfError(err)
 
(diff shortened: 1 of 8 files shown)


ankitpokhrel (commented):
Hi @martinpovolny, I like this idea but I think this needs a bit of work before we can merge.

The flag `fixed-columns` should be available and have same behavior across all interactive table listings. For instance, these commands should also behave in the same way as `jira issue list --fixed-columns 2`

```sh
$ jira epic list --table --fixed-columns 2
$ jira epic list <ID> --fixed-columns 2
$ jira sprint list --current --fixed-columns 2
$ jira sprint list <ID> --fixed-columns 2
```

_(Also, I rebased this branch)_

martinpovolny:
@ankitpokhrel : Updated as you requested.

Thx for the review!

ankitpokhrel:
Hi @martinpovolny thanks! 

I noticed slightly different behavior in explorer view for `jira epic list`, the `fixed-columns` option doesn't seem to work there. `jira sprint list` works as expected with `fixed-columns` tho. Could you please fix this inconsistency, looks good otherwise.

martinpovolny:
Yes! Good find! The epic explorer view did not respect the new option. 

e.g.
````
jira epic list --fixed-columns=2 --project APPSRE`
```

now works as expected.

ankitpokhrel (approved):
Looks good. Thank you @martinpovolny!

PR #517, opened 2022-12-03
Allow custom key mappings for the list of issues

martinpovolny:
With this, you can define custom keys and actions in the list of issues.

Example configuration (`~/.config/.jira`):

```
...
login: rhn-engineering-mpovolny
project:
  key: OSDEV
  type: ""
server: https://issues.redhat.com
custom_keys:
  issues:
    w: "jira issue view %KEY%"
    t: "&jira issue move %KEY% 'In Progress'; jira issue assign %KEY% rhn-engineering-mpovolny"
    s: '&jira sprint add $(jira sprint list --table --plain --no-headers --columns ID | head -1) %KEY%'
```

 * The "w" key is an example of a foreground running command mapped to the "w" key. It has no use, but demo.
 * The "t" key implements a "take" action -- issue is move to "In Progress" and then assigned to me. This is run in a separate goroutine (because of the `&` at the beginning of the command line).
 * The "s" key puts the issue into the latest sprint.
 
 This is just an example, this feature allows everyone to set his/her own mapping.
 
*This MR needs rebasing, but I'd like to open a discussion about the feature and the implementation.


ankitpokhrel:
Not sure if this would be the best way to add custom keys but it looks interesting. Let me play around a bit with this. Could you please resolve the conflicts and rebase. Thank you!

martinpovolny:
Rebased.

prateek:
One additional suggestion: It'd be great if this supported overriding the existing keybinds too - e.g. I'd prefer <enter> on `jira issue list` open the issue in the terminal and I'd define a bind like O or some such to open the issue in my browser.

PR #518, opened 2022-12-05
feat: Raise an error when invalid custom field is specified.

martinpovolny:
Raise an error when an invalid custom field is specified.

Currently, the invalid custom field is silently ignored which created a wrong impressing that everything is OK.


pkg/jira/create.go
```diff
@@ -230,6 +235,11 @@ func constructCustomFields(fields map[string]string, data *createRequest) {
 				data.Fields.M.customFields[configured.Key] = val
 			}
 		}
+
+		if !found {
+			fmt.Fprintf(os.Stderr, "\nInvalid custom field specified: %s\n", key)
+			os.Exit(1)
```
ankitpokhrel, line 241: Since this is a package, instead of exiting we should return appropriate error and let the caller decide what to do with the error.

ankitpokhrel (commented):
Hi @martinpovolny, thank you for the PR.

I think the validation was not added because the Jira API is flexible about this. The API simply ignores invalid fields and set the valid ones. The CLI behavior is same as this and I am not sure if we need to be strict on the CLI side. Also, this would be sort of a breaking change IMO. Thoughts?

pkg/jira/create.go
martinpovolny: Fixed.

martinpovolny:
>  The CLI behavior is same as this and I am not sure if we need to be strict on the CLI side. Also, this would be sort of a breaking change IMO. Thoughts?

The problem is that when you (as a user) have incorrectly set Custom fields in the configuration file, you have no idea, what is happening because the invalid custom field is silently ignored.

That is confusing and creates frustration when using the feature.

At the very least there should be a warning. You are passing an argument and expect it to have an effect. There's none.

I would not consider it a breaking change when the CLI newly checks this. It merely points out an error situation that was previously wrongly ignored.

ankitpokhrel:
> I would not consider it a breaking change when the CLI newly checks this. It merely points out an error situation that was previously wrongly ignored.

I think we should start with the warning and fail in future release. I would suggest the following:
- Add a warning when running the command so the users of this feature are aware of the incoming change.
- Do a new minor release with this change in place.
- A version after that will include the patch to fail for invalid custom field.

PR #544, opened 2023-01-02
feat(tui): Allow issue transition from the ui

ankitpokhrel:
Relates to #403 

With this change, pressing `m` in the TUI will let us quickly transition the issue status.

move-tui


muffinmad:
Cool feature!

Is it possible to display choices vertically? I have many statuses with some long names so not all the choices are visible.

Thanks!

PR #581, opened 2023-02-25
feat: Support for adding watcher to Issue

abhishekbiyala:


pkg/jira/issue.go
```diff
@@ -419,3 +419,37 @@ func (c *Client) RemoteLinkIssue(issueID, title, url string) error {
...
+		err  error
+		body []byte
+	)
+
+	body, err = json.Marshal(watcher)
+	if err != nil {
+		return err
+	}
+
+	res, err = c.Post(context.Background(), path, body, Header{
```
ankitpokhrel, line 438: We need to use v2 version of the API to make sure this works in older Jira versions.

```suggestion
	res, err = c.PostV2(context.Background(), path, body, Header{
```

pkg/jira/issue_test.go
```diff
@@ -554,3 +554,33 @@ func TestRemoteLinkIssue(t *testing.T) {
...
+
+func TestWatchIssue(t *testing.T) {
+	var unexpectedStatusCode bool
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "POST", r.Method)
+		assert.Equal(t, "application/json", r.Header.Get("Accept"))
+		assert.Equal(t, "application/json", r.Header.Get("Content-Type"))
+
+		assert.Equal(t, "/rest/api/3/issue/TEST-1/watchers", r.URL.Path)
```
ankitpokhrel, line 566: Let's make sure we are calling v2 version of the Jira API

```suggestion
		assert.Equal(t, "/rest/api/2/issue/TEST-1/watchers", r.URL.Path)
```

internal/cmd/issue/watch/watch.go
```diff
@@ -0,0 +1,301 @@
...
+				return errInvalidKeyword
+			}
+
+			return nil
+		},
+	}
+	return survey.Ask([]*survey.Question{qs}, &ac.params.user)
+}
+
+func (ac *watchCmd) searchAndWatchUser(project string) error {
```
ankitpokhrel, line 249: Nitpick: This method is not actually watching the user and I find the name bit confusing. Let's call it `searchAndSetUser`?

ankitpokhrel (commented):
Hi @abhishekbiyala I added some comments, looks good otherwise! Great work 🎉

ankitpokhrel:
@abhishekbiyala had to make some changes to make it work for older jira versions. Your commits are included and merged in #584. Thank you for your contribution 🎉

PR #608, opened 2023-04-27
dep: Upgrade cobra to v1.7.0

dbirks:
This fixes an issue with cobra, more of an annoyance really for me, where you couldn't easily source the zsh completion script like you could with bash.

Here's the upstream PR with more info, that was released recently with v1.7.0: https://github.com/spf13/cobra/pull/1917

Example of the completion not sourcing nicely before:

Screenshot from 2023-04-26 21-11-25

Completion after the cobra version bump:

Screenshot from 2023-04-26 21-09-55

---

Thanks a bunch for working on this CLI! I've only recently started using it, but have found it helpful.

Oh and that `make install` is quite handy. I didn't know you could `go install` from a local folder. Going to be using that a lot now. :slightly_smiling_face:


ankitpokhrel (approved):
Thank you @dbirks 🎉

PR #613, opened 2023-05-02
Add `browse_server` setting, allows for separate api/web client endpoints.

markhatch:
At my company, the on-prem installation has a different API endpoint from the web client endpoint.
This change allows "browse" URLs to point to a different base URL.

https://github.com/ankitpokhrel/jira-cli/discussions/604

API server: jira-api.example.com
UI server: jira.example.com

Changes:
Consolidate all browse URLs to use common GenerateServerBrowseURL function.
Add viper config option `browse_server` to be able to modify the generated browse URLs.

Tested by adding `browse_server` to `~/.config/.jira/.config.yml`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ bin/
 coverage/
 vendor/
 /.gocache
+
+.vscode/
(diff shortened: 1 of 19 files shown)


ankitpokhrel (approved):
This works great, thank you @markhatch 🎉

PR #615, opened 2023-05-07
feat: Add `mtls` authentication for client certificate auth

markhatch:
* Persist auth type in config file
* Update `jira init` to configure `mtls`
* Update README with instructions

Discussion: https://github.com/ankitpokhrel/jira-cli/discussions/603

```
$ jira init
? Installation type: Local
? Authentication type: mtls
? CA Certificate /home/mhatch/.certs/ca_list.pem
? MTLS Client Certificate /home/mhatch/.certs/mhatch.crt
? MTLS Client Key /home/mhatch/.certs/mhatch.key
? Link to Jira server: https://example.com/jira
? Login username: mhatch
? Default project: PROJECT
? Default board: PROJECT - Component work
```

### DIFF
--- a/internal/config/generator.go
+++ b/internal/config/generator.go
@@ -54,15 +54,24 @@ type issueTypeFieldConf struct {
 	}
 }
 
+// MTLS authtype specific config.
+type JiraCLIMTLSConfig struct {
+	CaCert     string
+	ClientCert string
+	ClientKey  string
+}
+
 // JiraCLIConfig is a Jira CLI config.
 type JiraCLIConfig struct {
 	Installation string
 	Server       string
+	AuthType     string
 	Login        string
 	Project      string
 	Board        string
 	Force        bool
 	Insecure     bool
+	MTLS         JiraCLIMTLSConfig
 }
 
 // JiraCLIConfigGenerator is a Jira CLI config generator.
@@ -81,6 +90,9 @@ type JiraCLIConfigGenerator struct {
 		epic         *jira.Epic
 		issueTypes   []*jira.IssueType
 		customFields []*issueTypeFieldConf
+		mtls         struct {
+			caCert, clientCert, clientKey string
+		}
 	}
 	jiraClient         *jira.Client
 	projectSuggestions []string
@@ -139,9 +151,23 @@ func (c *JiraCLIConfigGenerator) Generate() (string, error) {
 	if err := c.configureInstallationType(); err != nil {
 		return "", err
 	}
+
+	if c.value.installation == jira.InstallationTypeLocal {
+		if err := c.configureLocalAuthType(); err != nil {
+			return "", err
+		}
+	}
+
+	if c.value.authType == jira.AuthTypeMTLS {
+		if err := c.configureMTLS(); err != nil {
+			return "", err
+		}
+	}
+
 	if err := c.configureServerAndLoginDetails(); err != nil {
 		return "", err
 	}
+
 	if c.value.installation == jira.InstallationTypeLocal {
 		if err := c.configureServerMeta(c.value.server, c.value.login); err != nil {
 			return "", err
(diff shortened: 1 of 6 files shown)


markhatch:
@ankitpokhrel Here is a first pass at `mtls` auth. Let me know what you think!

internal/config/generator.go
```diff
@@ -190,6 +211,98 @@ func (c *JiraCLIConfigGenerator) configureInstallationType() error {
...
+		case strings.ToLower(jira.AuthTypeMTLS.String()):
+			c.value.authType = jira.AuthTypeMTLS
+		}
+	}
+
+	return nil
+}
+
+//nolint:gocyclo
+func (c *JiraCLIConfigGenerator) configureMTLS() error {
```
ankitpokhrel, line 246: Nit: I think we can get rid of `//nolint:gocyclo` with some minor refactoring

```go
func (c *JiraCLIConfigGenerator) configureMTLS() error {
	var qs []*survey.Question

	c.value.caCert = c.usrCfg.CaCert
	c.value.mtlsClientCert = c.usrCfg.MtlsClientCert
	c.value.mtlsClientKey = c.usrCfg.MtlsClientKey

	getIfEmpty := func(conf, name, msg, help string) {
		if conf != "" {
			return
		}
		qs = append(qs, &survey.Question{
			Name: name,
			Prompt: &survey.Input{
				Message: msg,
				Help:    help,
			},
		})
	}

	getIfEmpty(c.value.caCert, "cacert", "CA Certificate", "Local path to CA Certificate for your `server`")
	getIfEmpty(c.value.mtlsClientCert, "mtlsclientcert", "MTLS Client Certificate", "Local path to your client certificate")
	getIfEmpty(c.value.mtlsClientKey, "mtlsclientkey", "MTLS Client Key", "Local path to your client key")

	if len(qs) > 0 {
		ans := struct {
			CaCert         string
			MtlsClientCert string
			MtlsClientKey  string
		}{}

		if err := survey.Ask(qs, &ans); err != nil {
			return err
		}

		if ans.CaCert != "" {
			c.value.caCert = ans.CaCert
		}
		if ans.MtlsClientCert != "" {
			c.value.mtlsClientCert = ans.MtlsClientCert
		}
		if ans.MtlsClientKey != "" {
			c.value.mtlsClientKey = ans.MtlsClientKey
		}
	}

	return nil
}
```

ankitpokhrel (commented):
Hi @markhatch,

Apologies for late response and thank you for working on this!

I added some comments. Most of them are nitpicks. Looks good otherwise! Great Work 🎉

markhatch:
Think I've resolved most of the comments, let me know if it looks good!
Rebased to clean up the commits.

pkg/jira/client.go
```diff
@@ -93,14 +96,22 @@ func (e Errors) String() string {
...
-	AuthType AuthType
-	Insecure *bool
-	Debug    bool
+	Server     string
+	Login      string
+	APIToken   string
+	AuthType   AuthType
+	Insecure   *bool
+	Debug      bool
+	MTLSConfig MTLSConfig
```
markhatch, line 114: Wasn't sure if this should be a pointer instead?

internal/config/generator.go
```diff
@@ -190,6 +216,80 @@ func (c *JiraCLIConfigGenerator) configureInstallationType() error {
...
 }
 
+func (c *JiraCLIConfigGenerator) configureLocalAuthType() error {
+	var authType string
+
+	if c.usrCfg.AuthType == "" {
+		qs := &survey.Select{
+			Message: "Authentication type:",
+			Help:    "basic (login) or mtls (client certs)?",
+			Options: []string{"basic", "mtls"},
```
ankitpokhrel, line 226: Should here be `bearer` as well?

pkg/jira/client.go
ankitpokhrel: Should be ok even without pointer

ankitpokhrel (commented):
Hi @markhatch, apologies again for the late response. 

This looks good overall, just added few comments. I will try to get this merged for the next release. 

Thank you for working on this!

pkg/jira/client.go
```diff
@@ -226,7 +263,7 @@ func (c *Client) request(ctx context.Context, method, endpoint string, body []by
 
 	if c.authType == AuthTypeBearer {
 		req.Header.Add("Authorization", "Bearer "+c.token)
-	} else {
+	} else if c.authType == AuthTypeBasic {
```
ankitpokhrel, line 266: Current default auth type is basic (if nothing is set in config/env). Looks like this might break that?

internal/config/generator.go
markhatch: I believe that bearer is only used in for InstallationType cloud

pkg/jira/client.go
markhatch: Sorry been a while - but I believe this change was made so that the login/token are provided only if authType is AuthTypeBasic. 

If authType is Bearer you add the authorization header.
If authType is Basic you add the login/token
If authType is mtls, you don't need to modify the request.

I don't believe the default auth type has changed, and will depend on how the user configured authType.

ankitpokhrel (approved):
I wasn't able to test this fully but looks good overall. Thank you @markhatch

PR #619, opened 2023-05-09
feat: Enable issue edit to read body from stdin

erpel:
When working with `jira` I found that setting the body from stdin is not supported in `issue edit`. It seemed easy to add.

If you're interested in this feature that does not break anything (as far as I can tell) please feel free to accept this MR. I'm of course open to suggestions for anything that can be improved.


ankitpokhrel:
Hi @erpel, apologies for the late response.
 
I don't exactly remember the reason behind this check, TBH. Perhaps it was overlooked when working on the initial feature. I think we can update the method `StdinHasData` to something like the one below, so that it considers both stdin and file redirection.

```go
import "golang.org/x/crypto/ssh/terminal"

func StdinHasData() bool {
	if terminal.IsTerminal(int(os.Stdin.Fd())) {
		return false
	}
	return true
}

```

erpel:
Thanks for looking into this.
I took a look at "golang.org/x/crypto/ssh/terminal" and its IsTerminal function. Since it's just a wrapper around functionality in "golang.org/x/term" which was an indirect dependency already I opted to use this instead. Let me know what you think.

ankitpokhrel (approved):
This works fine. Thank you!

PR #624, opened 2023-05-24
Draft: support for multiple output formats

Miniwoffer:
Added a `--format` flag to `issue list` and `issue view` with supported formats:
 - plain ( same as `--plain`)
 - fancy (original format)
 - JSON

### DIFF
--- a/internal/cmd/issue/list/list.go
+++ b/internal/cmd/issue/list/list.go
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/ankitpokhrel/jira-cli/api"
+	"github.com/ankitpokhrel/jira-cli/internal/cmdutil/output"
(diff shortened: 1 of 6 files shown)


woodgear:
why close?

Miniwoffer:
Because i found this from the author after creating the MR
> @tstrijdhorst this CLI tool is created for a different purpose and adding json output would just complicate things. If you want a json output you can use tools like [httpie](https://github.com/httpie/httpie) or [curl](https://github.com/curl/curl) to call [the API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/) directly for a json response.
> 
> ```shell
> http GET https://<server>/rest/api/2/issue/ISSUE-1 --auth "<user>:${JIRA_API_TOKEN}"
> ```

Issue #587, opened 2023-03-10
Add support for Affected Versions (or just versions)


Roobxyz:
> Please consider [opening a discussion](https://github.com/ankitpokhrel/jira-cli/discussions/categories/ideas) in `ideas` category if applicable. Your feature will be prioritized on based on the number of votes.  

**Is your feature request related to a problem? Please describe.**
Affected Versions is a native JIRA field - so would be good to add support for it (not sure if we can reach it with custom fields)
https://www.atlassian.com/agile/tutorials/versions#:~:text=Affects%20version%20is%20the%20version,feature%20or%20bugfix%20to%20customers.

**Describe the solution you'd like**
To be able to supply `--affected-versions` during a `jira issue create`

**Describe alternatives you've considered**
Looked at custom fields but couldn't understand it

**Additional context**
A snippet from a request made directly to the API below shows `versions` is a first class field and is distinct to `fixVersions`
```
    "fixVersions": [],
    ...
    "timeestimate": null,
    "customfield_10218": null,
    "customfield_10219": null,
    "versions": [
      {
       "self":...
       "id":....
       "description":...
       "archived":....
       "released":.....
     }]
```

damianoneill:
@ankitpokhrel @Roobxyz its not called out explictly here but assume the same request for --fix-versions support for the native jira field Fix Versions as these are likely to be used at the same time in several scenarios.

damianoneill:
Ignore the comment above, just noticed the fix-version already exists as a command line option.

ankitpokhrel:
Done with #642

Roobxyz:
❤️

PR #642, opened 2023-07-07
feat: affects version

damianoneill:
@ankitpokhrel I cloned the code for fix-versions. As suggested in #587 the versions tag in the json payload could be used to set the affects version field.

I tested this for create issue only, against a cloud jira instance.

I'm not an expert in this API, (for example, in the fields.json testdata I'm unaware of this use, so not sure how/what to extend), so not in a position to say whether the PR is good or not, regardless, it should act as a starting point for this feature. 

Hope this helps,
Damian.


ankitpokhrel (approved):
This works as expected for edit and create in cloud & local instance. Thank you for taking care of this @damianoneill 🎉

Issue #593, opened 2023-03-24
How can I set custom fields on issue creation?


kwv:
### Discussed in https://github.com/ankitpokhrel/jira-cli/discussions/346

<div type='discussions-op-text'>

<sup>Originally posted by **ankitpokhrel** April 20, 2022</sup>
Jira supports a wide variety of datatypes when creating a custom field. Out of all [formats mentioned in the Jira doc](https://developer.atlassian.com/cloud/jira/service-desk/rest/intro/#fieldformats), the tool right now doesn't support `Cascading select custom field`, `Single-user picker custom field` and  `Multi-user picker custom field`. Support for these missing types will be added based on the request and the usecase.
</div>

----

For fields with unsupported/complex datatypes, would it be possible to allow the object representation to be passed in ?

```yml
    - name: Assigned PO
      key: customfield_10811
      schema:
        datatype: user
```
e.g. 
```console
jira issue edit EXMP-1234 --custom assigned-po='{"name": "$(jira me)"}'


Error:
  - customfield_10811: data was not an object

jira: Received unexpected response '400 '.
Please check the parameters you supplied and try again.
```

ankitpokhrel:
Hi @kwv, I am considering this to be an option, will check to see if its doable in the near future.

Roobxyz:
Hey - with my companies setup I'm unable to create Bug type issues since they require a custom field in the form of a Single-user picker custom field. 

I really like using this cli but this makes it difficult for me to use

GZLiew:
@ankitpokhrel is there a workaround for this? maybe I can edit the source 🤔

PR #661, opened 2023-09-06
Allow custom field values to be specified as JSON

octachrome:
This PR adds the ability to update custom fields whose type is not directly supported by `jira-cli`. This is achieved by specifying the JSON representation of the field value, e.g.:

```
jira issue create --custom 'json:found-in-version={"name":"Product 1.4"}'

jira issue edit ISSUE-123 --custom 'json:team={"set":"My Team"}' \
    --custom 'json:assigned-developer={"set":{"name":"MyUsername"}}'
```

If the field value starts with `json:`, the remaining string is interpreted as raw JSON. Note that when updating an issue, the `set` or `add` operation must also be provided, since this can be different for each custom field.

The implementation uses a new custom field struct type, `customFieldTypeJson`, and an implementation of `MarshallJSON` for that type which just returns the JSON stored in the struct.

After I implemented this I found #593 suggesting the same idea.

### DIFF
--- a/pkg/jira/create.go
+++ b/pkg/jira/create.go
@@ -228,9 +228,19 @@ func constructCustomFields(fields map[string]string, configuredFields []IssueTyp
 	data.Fields.M.customFields = make(customField)
 
 	for key, val := range fields {
+		rawJson := false
+		if strings.HasPrefix(key, "json:") {
+			key = key[5:]
+			rawJson = true
+		}
 		for _, configured := range configuredFields {
 			identifier := strings.ReplaceAll(strings.ToLower(strings.TrimSpace(configured.Name)), " ", "-")
 			if identifier != strings.ToLower(key) {
+				continue
+			}
+
+			if rawJson {
+				data.Fields.M.customFields[configured.Key] = customFieldTypeJson{Json: val}
 				continue
 			}
 
--- a/pkg/jira/customfield.go
+++ b/pkg/jira/customfield.go
@@ -39,3 +39,11 @@ type customFieldTypeProject struct {
 type customFieldTypeProjectSet struct {
 	Set customFieldTypeProject `json:"set"`
 }
+
+type customFieldTypeJson struct {
+	Json string
+}
+
+func (field customFieldTypeJson) MarshalJSON() ([]byte, error) {
+	return []byte(field.Json), nil
+}
--- a/internal/cmdcommon/create.go
+++ b/internal/cmdcommon/create.go
@@ -252,6 +252,9 @@ func ValidateCustomFields(fields map[string]string, configuredFields []jira.Issu
 
 	invalidCustomFields := make([]string, 0, len(fields))
 	for key := range fields {
+		if strings.HasPrefix(key, "json:") {
+			key = key[5:]
+		}
 		if _, ok := fieldsMap[key]; !ok {
 			invalidCustomFields = append(invalidCustomFields, key)
 		}
--- a/pkg/jira/edit.go
+++ b/pkg/jira/edit.go
@@ -361,9 +361,19 @@ func constructCustomFieldsForEdit(fields map[string]string, configuredFields []I
 	data.Update.M.customFields = make(customField)
 
 	for key, val := range fields {
+		rawJson := false
+		if strings.HasPrefix(key, "json:") {
+			key = key[5:]
+			rawJson = true
+		}
 		for _, configured := range configuredFields {
 			identifier := strings.ReplaceAll(strings.ToLower(strings.TrimSpace(configured.Name)), " ", "-")
 			if identifier != strings.ToLower(key) {
+				continue
+			}
+
+			if rawJson {
+				data.Update.M.customFields[configured.Key] = []customFieldTypeJson{{Json: val}}
 				continue
 			}
 


pkg/jira/create.go
```diff
@@ -228,12 +228,22 @@ func constructCustomFields(fields map[string]string, configuredFields []IssueTyp
 	data.Fields.M.customFields = make(customField)
 
 	for key, val := range fields {
+		rawJson := false
+		if strings.HasPrefix(key, "json:") {
+			key = key[5:]
+			rawJson = true
+		}
 		for _, configured := range configuredFields {
```
ankitpokhrel, line 236: Looks like we still need to go through configured field so why not just add a type called `json` that accepts the raw json data? 

For instance the configuration can be:
```yml
issue:
    fields:
        custom:
            - name: platform
              key: customfield_10032
              schema:
                datatype: json
```

And we can pass it as is

Screen Shot 2023-09-13 at 9 18 39 PM

Now, we don't need to pass `json:` tag?
```sh
jira issue create --custom='platform=[{"value":"Linux"}]'
```

ankitpokhrel (commented):
Hi @octachrome, Thank you for working on this.

I think we can get rid of `json:` prefix (unless I am missing something). Let me know what you think! Thanks again!

pkg/jira/create.go
octachrome: It is a good idea to have a `json` type that can be used in the config file. But is there an easy way to update the config file to set the type of custom fields? I think the Jira users on my team will find it hard to locate this file and make the right change by hand. Another inconvenience is that if a user runs `jira init` (for example if their password/access token changes), then the config file will be reset.

Looking at my config file, all the custom fields for which JSON is needed have type `any`. What do you think about using JSON by default for all fields of type `any`?

Another idea is to add an option to the `jira init` command to mark certain fields as needing JSON, e.g.:

```
jira init --json-custom-fields "first-field,second-field,etc"
```

ankitpokhrel: > But is there an easy way to update the config file to set the type of custom fields?

Not at the moment unfortunately. There is a plan to have something like [gh config](https://cli.github.com/manual/gh_config) but I don't have any timeline for it yet (open for PRs :)).

> I think the Jira users on my team will find it hard to locate this file and make the right change by hand.

Unless I am missing something, even with the `json:` tag you will still need to configure the field in config right? since we are looping through the configured fields in line#236?

```
for _, configured := range configuredFields {
```

> Looking at my config file, all the custom fields for which JSON is needed have type any. What do you think about using JSON by default for all fields of type any?

> Another idea is to add an option to the jira init command to mark certain fields as needing JSON, e.g.:

These are all valid ideas. The inconsistencies in Jira could make things a bit tricky so I am not sure if this will be the same case for all Jira installations.

I would suggest, let's start with simple and working implementation and we can make it better iteratively.

octachrome: > you will still need to configure the field in config right

Running `jira init` populates the config file with all custom fields, even the ones with unknown (`any`) type, so no manual config is necessary.

The current default behaviour for `any` type fields is to just send a string. I would rather not change that default behaviour in case it breaks something for an existing user (unless you know that this never happens in practice).

So I propose to add `json` as a new custom field type and add a flag to `jira init` to make it easy to configure these fields.

ankitpokhrel: Sounds good! Let's start by adding `json` as a new custom field type. We can talk about how we can make the config easily configurable.

octachrome: Here is my second attempt.

- Added `json` as a custom type, which can be used in the config file to force a field to use JSON when setting/editing.
- JSON is used by default for unknown field types, such as `user` or `version`.
- Factored out the code which constructs the `customFieldTypeXxx` objects so that it can be called recursively for arrays of any type, including JSON.
- Replaced the type-specific structs used to edit custom fields on issues, such as `customFieldTypeNumberSet` and `customFieldTypeStringSet` with generic structs that can be used to set/add/remove items on any field type.

I tested this on our Jira installation with custom fields of type `option`, `user`, `version`, `date`, `number`, `string`, `any` and array of `string`. This worked nicely, and I could specify `user` and `version` types as JSON from the command line with no manual changes to the config file needed.

JacobCallahan:
Would it be possible for the entire custom argument to be json?
`--custom '{"my field":"some value","another field":17}'`

octachrome:
> Would it be possible for the entire custom argument to be json? `--custom '{"my field":"some value","another field":17}'`

The disadvantage is that we would need to parse the JSON, separate out the fields, then reassemble the individual JSON pieces to send to Jira, which is awkward. What would be the benefit of this approach?

pkg/jira/create.go
octachrome: @ankitpokhrel Have you had a chance to review my update?

JacobCallahan:
There are two main advantages in my opinion.

1. The current proposal makes sense when considering that the value is json and not the field. However, the string syntax passed into custom is an awkward mix between non-json and json. It reads more like toml.
proposed:
`--custom='platform=[{"value":"Linux"}]'`

2. In order to use this, you have to edit pre-defined fields and set the type to json. This doesn't allow free-form data passed as truly custom fields in the request.

With this said, I fully acknowledge that the proposed change may not be possible with the proposed PR, hence the question.

Thanks for entertaining it!

octachrome:
> when considering that the value is json and not the field
I don't understand how anything except the value can be JSON. How can the field be JSON? Can you give an example?

> you have to edit pre-defined fields and set the type to json
I agree this is the main disadvantage, and your suggestion does overcome that with a nicer syntax than what I originally proposed.

> This doesn't allow free-form data passed as truly custom fields in the request.
Please give examples of what you mean by "free-form data" and "truly custom fields".

I am using the Jira CLI to automate actions in shell scripts, and I expect other people are too. Assembling a single JSON object to describe all my custom field values is not easy with a shell script, keeping track of where commas are needed, etc. An extension of your suggestion is to allow the option to be passed more than once. So both of the following would be equivalent.

    --custom-json '{"field1": "value1", "field2": "value2"}'
    --custom-json '{"field1": "value1"} --custom-json '{"field2": "value2"}'

I would also use a different option name to avoid confusion with `--custom`.

pkg/jira/create.go
ankitpokhrel: Hi @octachrome, sorry for the late response. Would you be able to provide some examples on how to construct valid requests for different types of custom fields? Thanks!

abhisheksoni27:
any update on this one? this will make things much easier when transitioning issues, updating custom fields etc. @ankitpokhrel @octachrome

pkg/jira/create.go
octachrome: @ankitpokhrel, if you mean CLI examples, see below (the syntax is the same for the `issue create` command):

```
# Field of type "version"
jira issue edit --no-input --custom 'found-in-version={"name": "MyProduct 1.0"}' XYZ-1234
# Field of type "user"
jira issue edit --no-input --custom 'assigned-developer={"name": "BobFossil"}' XYZ-1234
```

I could add these to the readme, or it might be better to add them to the article here: https://github.com/ankitpokhrel/jira-cli/discussions/346.

josete89:
Is there any plan to merge this at any point?

ankitpokhrel:
Sorry I got distracted with other priorities for a while now. I'll check if this can be merged soon.

pkg/jira/customfield.go
```diff
@@ -1,41 +1,88 @@
...
+	switch dataType {
+	case customFieldFormatOption:
+		return customFieldTypeOption{Value: value}
+	case customFieldFormatProject:
+		return customFieldTypeProject{Value: value}
+	case customFieldFormatArray:
+		pieces := strings.Split(strings.TrimSpace(value), ",")
+		items := make([]any, len(pieces))
+		for idx, piece := range pieces {
+			items[idx] = constructCustomField(itemType, piece, "")
```
ankitpokhrel, line 60:
```suggestion
			items[idx] = constructCustomField(itemType, "", piece)
```


We seem to be setting empty value here. That breaks the following:

```sh
$ jira issue create -s"Some story" --custom platform=Linux --debug
```

caseycoding:
I've verified this update works for my use case.  Thank you for this work!

evg656e:
I’ve tested this as well, and it works like a charm. I hope to see it merged upstream soon.

snoby:
Jira really has made this utterly too complex.  Anyway to do something like this?
```
jq .fields.customfield_11800
{
  "self": "https://jira-eng-gpk2.funcompany.com/jira/rest/api/2/customFieldOption/14008",
  "value": "Analytics",
  "id": "14008",
  "disabled": false,
  "child": {
    "self": "https://jira-eng-gpk2.funcompany.com/jira/rest/api/2/customFieldOption/42201",
    "value": "Team fun  2.0",
    "id": "42201",
    "disabled": false
  }
}

```

octachrome:
> Jira really has made this utterly too complex. Anyway to do something like this?

I looked at fields like this recently and I couldn't figure out how to get them working. Way too complicated.

PR #662, opened 2023-09-13
add command line arg delimiter for custom delimiter instead of tab.

iksi4prs:
**Motivation**
Adding support to parse the output of issues in plain mode,

**Current problem**
Current delimiter in plain mode is TAB (\t), but TAB is also used for padding the columns.
If the an issue has a value which requires more than one TAB for padding, eg where issue type is of "improvement",
then some columns will contain more than one TAB, hence not possible to parse in simple way.

**Change**:
Add command line arg delimiter for custom delimiter instead of tab supported only for plain mode for issues.
For backward compatibility and reduce risk of affecting other code, this new command line arg is applicable only for plain mode.


ankitpokhrel:
Hi @iksi4prs, thank you for working on this. Would you be able to fix/update the tests? Thanks

iksi4prs:
@ankitpokhrel , will do my best. There are few old opened PRs (from other people), hence I thought this project is not active anymore.

jcutrell:
Is this still active? I'd really like to have the delimiter option to do some analysis on the issue list.

iksi4prs:
@jcutrell I will do my best to fix the tests asap, so pr can be merged. 
in any case, u can take my branch and build it, the feature itself works,
it just that uni-tests were not updated.

iksi4prs:
@ankitpokhrel , I've fixed the unit tests and add new one for this feature. please check when u have time. 10x.

ankitpokhrel:
Sorry for getting into this late

PR #669, opened 2023-10-16
Add support for updating the estimate

chapmanc:
Referencing these docs to update the estimate when adding a worklog: 

https://docs.atlassian.com/software/jira/docs/api/REST/9.11.0/#api/2/issue-addWorklog

I've added tests but would like to do an e2d test as well.


ankitpokhrel (approved):
This seems to be working great. Thank you @chapmanc 🎉

PR #702, opened 2024-01-18
Bugfix: Properly close created files

beatbrot:
I noticed the tests failing on my windows machine.

So I searched for the issue and found that created files are never closed. The fix is small and simple :)


ankitpokhrel (approved):
Thank you @beatbrot

PR #706, opened 2024-02-04
Consider Windows Terminal as non-dumb terminal

beatbrot:
This is in accordance to [this comment](https://github.com/microsoft/terminal/issues/1040#issuecomment-496691842). As discussed further in the issue, in an ideal world, we wouldn't check "user agents" of terminals one-by-one, but instead would do proper feature probing like described [here](https://github.com/termstandard/colors).

But anyway: This is a good first step to enhance Windows support.

### DIFF
--- a/pkg/tui/helper.go
+++ b/pkg/tui/helper.go
@@ -64,13 +64,14 @@ func getActionModal() *primitive.ActionModal {
 		SetTextColor(tcell.ColorDefault)
 }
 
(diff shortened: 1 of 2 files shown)


beatbrot:
@ankitpokhrel Could you maybe take another look at this? :) Having this fix would really improve my usecase :)

ankitpokhrel (approved):
I don't have a windows machine to test this at the moment but looks good. Thanks!

JosXa:
Waiting desperately for a release of this :)

Issue #718, opened 2024-03-12
JSON output option for `issue view`


mpalahuta:
**Is your feature request related to a problem? Please describe.**
Hey! Thanks a lot for this great tool! I want to use it in a release automation project for a client company. However, I'll need to be able to format the output of the `jira issue view` in a machine-readable way (JSON). Additionally, I'll also need to add the `emailAddress` field to the issue data structure. Would that be OK, @ankitpokhrel? 

Use case: Jira automation from Jenkins pipeline scripts; using email to link users between various systems f.e. Slack.

**Describe the solution you'd like**
I'd like to create two PRs:
- A PR adding the `emailAddress` field to the `pkg/jira/IssueFields` data structure and update the `jira view issue` to display the field as well (for consistency) https://github.com/ankitpokhrel/jira-cli/pull/719
- A PR adding `--json` option to the `jira view issue` command that would make the command print the output as plain JSON if the flag is set https://github.com/ankitpokhrel/jira-cli/pull/720

**Describe alternatives you've considered**
Parsing the current output programmatically is hard and not reliable

ankitpokhrel:
Closed with #720

PR #719, opened 2024-03-12
feat(718): view issue: assignee and reporter email

mpalahuta:
https://github.com/ankitpokhrel/jira-cli/issues/718

This change only applies to the `jira issue view <key>` command. Parsing and displaying email addresses of the assignee and the reporter of the viewed issue if the emails are present in the Jira API response.

Screenshot 2024-03-12 at 16 14 32

### DIFF
--- a/internal/view/epic_test.go
+++ b/internal/view/epic_test.go
@@ -19,13 +19,15 @@ func TestEpicData(t *testing.T) {
 			}{Name: "Fixed"},
 			IssueType: jira.IssueType{Name: "Epic"},
 			Assignee: struct {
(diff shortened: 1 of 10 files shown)


ankitpokhrel:
Hi @mpalahuta, could you please help me understand the usecase for this? In the issue report you mentioned this is for consistency. Do you mean to be consistent with what we are parsing from the API response?

mpalahuta:
Yep, exactly! However, since we’re gonna output the raw API response this PR is no longer relevant, closing it. Thanks!

PR #720, opened 2024-03-14
feat(718): issue view: raw response option

mpalahuta:
https://github.com/ankitpokhrel/jira-cli/issues/718

Adding `--json` flag to the `jira issue view <KEY>` command so that the tool prints the issue contents as plain JSON if the flag is set. This feature allows users to process the issue information with their programs to support various automation scenarios.

I've added an intermediate struct `printableIssue` to ensure correct null/empty values in the produced JSON, to avoid zero-values like 
```
{
...
  "assignee": {
    "displayName": ""
  }
...
}
```

Screenshot 2024-03-14 at 16 29 36

### DIFF
--- a/internal/cmd/issue/view/view.go
+++ b/internal/cmd/issue/view/view.go
@@ -1,6 +1,8 @@
 package view
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
@@ -16,7 +18,20 @@ const (
 	examples = `$ jira issue view ISSUE-1
 
 # Show 5 recent comments when viewing the issue
-$ jira issue view ISSUE-1 --comments 5`
+$ jira issue view ISSUE-1 --comments 5
+
+# Get the raw JSON data
+$ jira issue view ISSUE-1 --raw`
+
+	flagRaw      = "raw"
+	flagDebug    = "debug"
+	flagComments = "comments"
+	flagPlain    = "plain"
+
+	configProject = "project.key"
+	configServer  = "server"
+
+	messageFetchingData = "Fetching issue details..."
 )
 
 // NewCmdView is a view command.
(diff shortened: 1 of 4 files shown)


internal/view/issue.go
```diff
@@ -457,3 +458,149 @@ func (i Issue) renderPlain(w io.Writer) error {
...
 }
+
+// printableIssue is a jira.Issue that can be marshaled into a format of choice f.e. JSON and
+// printed for other programs to consume.
+type printableIssue struct {
+	Key    string               `json:"key"`
+	Fields printableIssueFields `json:"fields"`
+}
+
+type printableIssueFields struct {
```
ankitpokhrel, line 469: Instead of sending few fields, why not output the exact response from the API? I am pretty sure there are usecases that require fields that we haven't considered in the response.

ankitpokhrel (commented):
Hi @mpalahuta, Thank you for working on this. 

I was bit hesitant to support json response but since there are [lot of requests](https://github.com/ankitpokhrel/jira-cli/discussions?discussions_q=is%3Aopen+json) for this (and now a PR), I am reconsidering my decision.

Having said that, if we want to go ahead with json response, I think we should send the entire raw json response from the API to the user. That would help us cover most of the usecase reported.

Let me know what you think!

mpalahuta:
Hey @ankitpokhrel 
Thanks a lot for looking into this PR and for the green light on the JSON output - this will be very helpful!

Reg. raw response, sounds great, thanks! I will update this PR on Mon to do exactly that!

mpalahuta:
hey @ankitpokhrel 
I've updated the PR to follow your suggestion of printing the raw API response. Verified on my machine - works as expected.
Please take a look whenever you have a chance, thanks!

Screenshot 2024-03-19 at 18 01 24

valters:
hi @ankitpokhrel, I am also interested in this improvement. Having this will enable new build/release automation scenarios for us.

internal/cmd/issue/view/view.go
```diff
@@ -16,7 +18,20 @@ const (
...
-$ jira issue view ISSUE-1 --comments 5`
+$ jira issue view ISSUE-1 --comments 5
+
+# Get the raw JSON data
+$ jira issue view ISSUE-1 --raw`
+
+	flagRaw      = "raw"
+	flagDebug    = "debug"
+	flagComments = "comments"
+	flagPlain    = "plain"
```
ankitpokhrel, line 29: Thank you for refactoring this 👍🏻

ankitpokhrel (approved):
This looks great! Thank you 🎉

PR #741, opened 2024-05-23
Allow search for issues of all projects

beatbrot:
This is a very quick & rough implementation that allows us to
search for issues in all projects by passing an empty string as
project.

It effectively allows us to run `jira issue list -p ""` and get issues of all projects:

```
KEY     SUMMARY
BB-1    My fist story
AA-9    As a developer, I'd like to update story status during the sprint
```


ankitpokhrel:
This can be done using raw JQL as mentioned in [the FAQ](https://github.com/ankitpokhrel/jira-cli/discussions/420).

PR #756, opened 2024-08-06
[sprint] Add cli method for close sprint with supporting jira client …

cworsnup13:
Add cli method for close sprint with supporting jira client methods. Sprint update is using full update rather than partial. Default behavior is that open tickets move to backlog. 

Successful failure to update when sprint is in closed state.


ankitpokhrel (approved):
Thank you!  Seems to be working fine 🎉  I'll make some small changes in subsequent PRs.

PR #748, opened 2024-07-11
Feature: add original estimate option to issue creation

rodcloutier:
This PR intends to add the possibility of specifying the original estimate when creating an issue. 

* New option is called `-e, --original-estimate`
* Will set the `{ timetracking: { originalEstimate: "" }` value in the payload.

```shell 
❯ ./jira issue create --help
Create an issue in a given project with minimal information.

USAGE
  jira issue create [flags]

FLAGS
  -t, --type string                   Issue type
  -P, --parent string                 Parent issue key can be used to attach epic to an issue.
                                      And, this field is mandatory when creating a sub-task.
  -s, --summary string                Issue summary or title
  -b, --body string                   Issue description
  -y, --priority string               Issue priority
  -r, --reporter string               Issue reporter (username, email or display name)
  -a, --assignee string               Issue assignee (username, email or display name)
  -l, --label stringArray             Issue labels
  -C, --component stringArray         Issue components
      --fix-version stringArray       Release info (fixVersions)
      --affects-version stringArray   Release info (affectsVersions)
  -e, --original-estimate string      Issue Original estimate
      --custom stringToString         Set custom fields (default [])
  -T, --template string               Path to a file to read body/description from
      --web                           Open in web browser after successful creation
      --no-input                      Disable prompt for non-required fields
  -h, --help                          help for create

INHERITED FLAGS
  -c, --config string    Config file (default is /Users/rcloutier/.config/.jira/.config.yml)
      --debug            Turn on debug output
  -p, --project string   Jira project to look into (defaults to /Users/rcloutier/.config/.jira/.config.yml)

EXAMPLES
  $ jira issue create

  # Create issue in the configured project
  $ jira issue create -tBug -s"New Bug" -yHigh -lbug -lurgent -b"Bug description"

  # Create issue in another project
  $ jira issue create -pPRJ -tBug -yHigh -s"New Bug" -b$'Bug description\n\nSome more text'

  # Create issue and set custom fields
  # See https://github.com/ankitpokhrel/jira-cli/discussions/346
  $ jira issue create -tStory -s"Issue with custom fields" --custom story-points=3

  # Load description from template file
  $ jira issue create --template /path/to/template.tmpl

  # Get description from standard input
  $ jira issue create --template -

  # Or, use pipe to read input directly from standard input
  $ echo "Description from stdin" | jira issue create -s"Summary" -tTask

  # For issue description, the flag --body/-b takes precedence over the --template flag
  # The example below will add "Body from flag" as an issue description
[...]

PR #767, opened 2024-08-26
fix: regression introduced by new OriginalEstimate attribute

fabio42:
The recent changes in PR #748 caused issues with jira-cli when used with Jira Enterprise On-Prem (now referred to as DataCenter). All requests started failing with the following error message:

```
Error:
  - timetracking: Field 'timetracking' cannot be set. It is not on the appropriate screen, or unknown.
```

The issue arises because, even if the `--original-estimate` option is not specified, the timetracking field is included in the JSON query. This PR addresses the issue by ensuring that the timetracking field is only included in the JSON query if the `--original-estimate` option is set. I have tested the fix, and all tests are passing.

Please review the changes and let me know if there are any concerns.

### DIFF
--- a/pkg/jira/create.go
+++ b/pkg/jira/create.go
@@ -133,9 +133,6 @@ func (*Client) getRequestData(req *CreateRequest) *createRequest {
 		Summary:   req.Summary,
 		Labels:    req.Labels,
 		epicField: req.EpicField,
-		TimeTracking: struct {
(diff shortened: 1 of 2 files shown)


itay-sho:
Thanks for the fix! had the same problem

ankitpokhrel (approved):
Thank you @fabio42

PR #768, opened 2024-08-27
fix: timetracking field always marshalled

wbh1:
Currently, the new field from https://github.com/ankitpokhrel/jira-cli/pull/748 for timetracking is always marshalled to JSON even if the "OriginalEstimate" field is an empty string.

This is because of the way the field is initialized as a struct that will never be nil and therefore never caught by the 'omitempty' tag.

This changes the TimeTracking field to be a pointer, which is nil by default, so that if OriginalEstimate isn't set, the timetracking field won't be in the JSON sent to Jira.

This is a problem if the issue type being created is one that doesn't support a timetracking field.

### DIFF
--- a/pkg/jira/create.go
+++ b/pkg/jira/create.go
@@ -133,9 +133,6 @@ func (*Client) getRequestData(req *CreateRequest) *createRequest {
 		Summary:   req.Summary,
 		Labels:    req.Labels,
 		epicField: req.EpicField,
-		TimeTracking: struct {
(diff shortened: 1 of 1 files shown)


wbh1:
I will look at fixing tests soon.

ankitpokhrel:
This seems to be fixed by https://github.com/ankitpokhrel/jira-cli/pull/767

Thank you @wbh1

wbh1:
🤦‍♂️  I should've checked open PRs. Thanks!

PR #769, opened 2024-08-27
feat: add easy issue search to list cmd

wbh1:
Adds the ability to provide an optional positional arg to search for text in issues.

It functions the same as searching via the search bar in the Jira web UI and is appended to any raw JQL query provided in the "-q" flag.

I'm also open to refactoring this into its own subcommand, if that's preferable. I know this is already possible with the "-q" flag, this is just intended to be a nice convenience function.


ankitpokhrel (approved):
This looks good, thank you for contributing and sorry for the late review.

PR #775, opened 2024-09-16
feat: Add option to output issue creation result as Json

sebbbastien:
To be able to use `jira-cli` in an automation context, issue creation output should be in JSON.

I don't know if `--json` is a good choice, as `--raw` is already used for issue view. Maybe I have to change `--json` to `--raw`.


ankitpokhrel:
Hi @sebbbastien, thank you for the PR and sorry for late response. Could you pls update `--json` to `--raw` for consistency reasons?

mghesh-yseop:
Hi,
I love the idea of `json` output. Could it be generalized to all `--raw` outputs ?

ankitpokhrel (approved):
Works great! Thank you. And sorry again for late response!

ankitpokhrel:
> Could it be generalized to all commands output ?
I think that's a fair ask given that the tool is mostly being used in CI/CD pipeline. I'll see what can be done. Contributions are always welcome :)

PR #793, opened 2024-11-05
fix: case insensitive sub-task comparison

tiemma:
Within my cloud installation, the jira init command pulls the handle and name just fine but my administrator configured the name as "Sub-task" and the handle as "Sub-Task" which fails the equal check and causes failures with the parent key assignment. 

```
- id: "10093"
  name: Sub-task
  handle: Sub-Task
  subtask: true
```

This PR addresses this by using an EqualFold so case insensitive matching is done for the issue type check.


ankitpokhrel (approved):
Thank you!

PR #811, opened 2024-12-31
Allowed multiple replace strings in issue clone command

AdamSaleh:
As I use some of my jira issues as templates for new issues, I realized when scripting that having the ability to use multiple -H flags in *jira issue clone* command would be really useful.

The changes to the code to suport this were fairly minimal, and as far as I understand the configuration of flags, it shouldn't break any of the existing usage.

I will gladly include any changes necessary to have this merged :-)


ankitpokhrel (approved):
This works great! Thank you and sorry for late review.

PR #828, opened 2025-03-10
fix: use `tview.Escape` with standard square brackets

rpadaki:
The original implementation of `prepareTitle`, as well as the date ranges for sprints, used nonstandard brackets `⦗⦘` as replacements for square brackets `[]` in a hack to work around `tview` interpreting square brackets as format tags. This was causing visual problems for some powerline fonts.

The [documentation](https://pkg.go.dev/github.com/rivo/tview) for `tview` mentions that bracketed strings may be escaped by typing `[text[]` instead of `[text]`, and the package provides `tview.Escape(str)` to perform that conversion.

This change switches to using the escape function. I tried to minimize the amount of things I was touching, which also means that the expected strings in tests are a bit ugly.

### DIFF
--- a/internal/view/board_test.go
+++ b/internal/view/board_test.go
@@ -14,7 +14,7 @@ func TestBoardRender(t *testing.T) {
 
 	data := []*jira.Board{
 		{ID: 1, Name: "First", Type: "scrum"},
-		{ID: 2, Name: "⦗2⦘ Second", Type: "kanban"},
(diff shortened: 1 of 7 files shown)


rpadaki:
(I'm getting CI failures due to pre-existing linting errors in some of the test files. I'd prefer not to address those in this PR.)

ankitpokhrel (approved):
Seems to work great. Thank you!

PR #851, opened 2025-05-23
feat: Add releases Subcommand

vkhitrin:
Add a subcommand for interacting with Jira releases (project versions)
using the non paginated API endpoint (which provides less
information compared to the paginated endpoint).


ankitpokhrel (approved):
LGTM overall. I'll do some minor updates later.

Thank you for contributing!

Issue #713, opened 2024-03-04
Support JIRA_CONFG env variable


weshouman:
**Is your feature request related to a problem? Please describe.**
Managing multiple boards (or even multiple servers) would require setting the configuration flag every time.

**Describe the solution you'd like**
Exposing `JIRA_CONFIG` env variable which if set, the configuration file in it would be used.

**Describe alternatives you've considered**
Setting the `-c config-file directly`

**Additional context**
It's possible to define a bash function that wrap the command like the following
but it would be much cleaner if that's supported natively
```bash
jira() {
    if [ -n "$JIRA_CONFIG" ]; then
        # JIRA_CONFIG is set, use it with --config
        command jira --config "${JIRA_CONFIG}" "$@"
    else
        # JIRA_CONFIG is not set, call jira normally
        command jira "$@"
    fi
}
```

weshouman:
Hello @ankitpokhrel, `XDG_CONFIG_HOME` would require having multiple dirs to have multiple configurations, while the `-c` flag allows having multiple configurations even in the same directory, I currently work around that by the provided snippet.

Note: may environment includes something like
```
$CONFIG_DIR/.jira/config1.config.yaml
$CONFIG_DIR/.jira/config2.config.yaml
...
```

AutomationD:
My usecase for this is to have jira/repo project mapping in the repo .envrc. `XDG_CONFIG_HOME` wouldn't work as it would set config directories for all other tools (such as podman)

dimmkirr:
Funny enough, I'm coming back to this for setting up a Jira per-tmux session, but seems like it still doesn't work.

pataquets:
Although feature is mentioned on readme, I've pulling my hair untill I've realized it's not in any release yet.
@ankitpokhrel: Fancy to cut a release?

Issue #434, opened 2022-08-13
Support for multiple Jira instances.


wstewarttennes:
**Is your feature request related to a problem? Please describe.**

I work with multiple Jira instances for work/personal life. Would LOVE to be able to access them all from either one jira-cli installation or across multiple iTerm profiles. Right now, because there is only one configuration file, I can't setup multiple Jira instances.

**Describe the solution you'd like**

option 1: Add support for configuration files with multiple Jira instances or pointing to multiple .netrc files. Then, add a JIRA_INSTANCE environment variable that I could set at launch that defines which jira instance I'm using.

options 2: Add support for configuration files with multiple Jira instances and add a flag in the cli that specifies which instance to use.

ankitpokhrel:
Hi @wstewarttennes, there is a `--config/-c` flag available that you can use to target different Jira instances. You can also use `XDG_CONFIG_HOME` env for the same purpose.

PR #858, opened 2025-05-26
Add support for environment variable to specify config file path

aazon:
Solves #713 #434

This pull request introduces a new environment variable, `JIRA_CONFIG_FILE`, which allows users to specify the path to the configuration file.

A common use case is working with multiple Jira servers across different projects simultaneously. In such scenarios, it’s helpful to maintain separate configuration files to avoid conflicts, especially with credentials.

While the solution proposed in #434 — using the `--config/-c` flag — is a valid option, it requires specifying the argument manually for every command. This approach is prone to errors and adds complexity when writing automated scripts.

Similarly, the solution proposed in #713 — relying on the `XDG_CONFIG_HOME` environment variable — affects a broader range of tools beyond `jira-cli`, which may lead to unintended side effects.

Using a dedicated environment variable like `JIRA_CONFIG_FILE` offers a more isolated, portable, and context-specific solution. It integrates cleanly into different workflows without impacting other tools or requiring repeated CLI arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,15 @@ default.
 #### Shell completion
 Check `jira completion --help` for more info on setting up a bash/zsh shell completion.
 
+#### Multiple projects
+
+The tool honors the `JIRA_CONFIG_FILE` environment variable, which specifies the location of the configuration file:
+
+```sh
+JIRA_CONFIG_FILE=./local_jira_config.yaml jira issue list 
+
+```
+
 ## Usage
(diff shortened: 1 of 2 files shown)


ankitpokhrel (approved):
Hi @aazon, Thank you for working on this!

mawid6:
oh, wait, sorry. 
```
jira version
(Version="1.6.0", GitCommit="6a6611ce59f1a85203b021af9113b2a4e82ecaef", CommitDate="2025-04-19T09:33:11+00:00", GoVersion="go1.24.1", Compiler="gc", Platform="windows/amd64")
```

Duh, this has not been packaged yet, I see. Apologies for the noise.

PR #877, opened 2025-07-17
feat: Enable OAuth 3LO support

### LINKED ISSUE #863: Add new JIRA_AUTH_TYPE OAuth 2.0 3LO
> Please consider [opening a discussion](https://github.com/ankitpokhrel/jira-cli/discussions/categories/ideas) in `ideas` category if applicable. Your feature will be prioritized on based on the number of votes.  

**Is your feature request related to a problem? Please describe.**
Company has migrated to Cloud and has disabled creating `Access Tokens` for security reasons. Hence, I'm unable to use this CLI. 

**Describe the solution you'd like**
Support OAuth 2.0 3LO: https://developer.atlassian.com/cloud/oauth/getting-started/enabling-oauth-3lo/

**Describe alternatives you've considered**

**Additional context**

With support for OAuth 2.0, users can continue to use this CLI and ensure that we've a good user-base for this project.

christianarty:
## Summary

fixes #863

This PR allows for users to have another option (`oauth`) when generating their JIRA `config.yml` for the Cloud installation. 

## Details
This PR implements [JIRA's 3LO OAuth](https://developer.atlassian.com/cloud/oauth/getting-started/enabling-oauth-3lo/) solution for users to obtain a JIRA access token. 
Each consumer of `jira-cli` will need to create a JIRA App with the specific scopes in order to connect it properly with their JIRA cloud instance. 

The oauth secret will be stored in the `.config/.jira` directory, where the tokens will be automatically regenerated when it expires and the newly generated tokens will be cached to the oauth secret file. 

Ideally, for OAuth, we would have one single distributed app that can be installed in multiple different JIRA cloud instances. However, The 3LO doesn't support [Proof Key for Code Exchange (PKCE)](https://oauth.net/2/pkce/). Without this support, we would have to share the single distrubuted app's client secret with all the consumers. See these links for more info: 

  - https://jira.atlassian.com/browse/ECO-283
  - https://community.developer.atlassian.com/t/oauth-2-0-with-proof-key-for-code-exchange-pkce/80173/3

As noted in the forum above, a workaround would be that each consumer has to create their own JIRA app and use that app's client ID and secret in the `jira-cli` client app.
  - This basically acts like a proxy to funnel requests into your JIRA cloud instance

## Testing Done
- `make deps install` => WORKS
- `~/go/bin/jira issue create -tTask -s"TEST TICKET" -l"testing" --template ~/jira/task.tmpl -a$(~/go/bin/jira me)` => WORKS (created a ticket, and proper link)

### Commits
- **(build) Update Golang OAuth2 Library to latest**
- **(feat) Implement OAuth2 func. into jira-cli**
- **feat:(oauth) - Ability to get cloud id for clients**
- **(docs) Add note about OAuth weirdness**
- **(refactor) Move oauth to separate module**
- **(refactor) Remove unnecessary args from `verifyLoginDetails` and `configureServerMeta`**
- **(tests): Add tests for oauth**
- **(nit) Move getToken logic into separate func**
- **(refactor) Organizing the oauth to break up the logic to other files**
- **(feat) Add the ability to load the persisted oauth token from filesystem**
- **(test) More tests for oauth**
- **(feat) Enable auto refreshing access tokens through the oauth2 custom transport in the client**
- **(nit) remove dead code**
- **(fix) The constructed server urls were pointing to authorization server, rather than their own JIRA server**
- **(docs) update README with more instructions.**


shadyabhi:
cc: @ankitpokhrel for visibility. It'll be great if this can be reviewed/merged soon, so folks that can only authenticate through `OAuth 3LO` can start using this tool again.

christianarty:
@ankitpokhrel Could you take a look at this PR when you get a chance? I think this is a nice feature add (and personally been using it for months now without hiccup)

ankitpokhrel:
@christianarty Thanks for the PR! I'm currently away, I'll look into this in a few weeks.

christianarty:
> I have a suggestion here.
> 
> I played with the app and 3LO.
> 
> It works great except when the app you created doesn't have the expected scope
> 
> I feel like the CLI should display the scope it expect in the console. Either by default, or when the timeout happens.
> 
> The error message on the consent screen provided by Jira is unclear.
> 
> Also, I'm unsure if you can report the scope the app was created with without being authenticated, but it would be great to double check what us expected and what is currently on the created oauth app

@ccoVeille Great suggestion! I am not sure about the scope when unauthorized and I think that should be a separate PR (this PR is already big enough tbh), but I did just add for the CLI to print out the expected scopes when creating a new config. It looks like this:
Screenshot 2025-11-16 at 12 51 05 PM

Do you think it should look like that or another format?
EDIT: I removed the `offline_access` scope from printing to console since that is a special scope you don't have to click on JIRA ui.

@ankitpokhrel I was **just a bit off ** when I said a couple days I'll fix it 😅  but I refactored this to use the keyring storage as the primary and file system as secondary and it works as expected. Also wrote some tests to account for it but I think this would suffice. I also updated the README to account for some of the changes I did with the env vars.

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,452 @@
...
+
+	"github.com/ankitpokhrel/jira-cli/internal/cmdutil"
+	"github.com/ankitpokhrel/jira-cli/pkg/utils"
+)
+
+const (
+	// JIRA OAuth2 endpoints.
+	jiraAuthURL            = "https://auth.atlassian.com/authorize"
+	jiraTokenURL           = "https://auth.atlassian.com/oauth/token"
+	accessibleResourcesURL = "https://api.atlassian.com/oauth/token/accessible-resources"
```
ccoVeille: Make use the same prefix as above


```suggestion
	jiraAccessibleResourcesURL = "https://api.atlassian.com/oauth/token/accessible-resources"
```

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,452 @@
...
+		Help:    "The redirect URL for Jira App. Recommended to set as localhost.",
+	}
+
+	if envClientID == "" {
+		if err := survey.AskOne(q1, &answers.ClientID); err != nil {
+			return nil, err
+		}
+	} else {
+		answers.ClientID = envClientID
+	}
```
ccoVeille: I would have written this to be simpler


```suggestion
	answers.ClientID = envClientID
	if envClientID == "" {
		if err := survey.AskOne(q1, &answers.ClientID); err != nil {
			return nil, err
		}
	}
```

Same for other blocks

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,452 @@
...
+
+// performOAuthFlow executes the OAuth authorization flow.
+func performOAuthFlow(config *OAuthConfig, httpTimeout time.Duration, openBrowser bool) (*oauth2.Token, error) {
+	fmt.Printf("Expected Scopes:\n")
+	for i, scope := range config.Scopes {
+		// Offline_access is a special scope to request a refresh token that wont be there to display in the JIRA UI, so we skip it.
+		if scope != "offline_access" {
+			fmt.Printf("  %2d. %s\n", i+1, scope)
+		}
+	}
```
ccoVeille: I tend to write something like this


```suggestion
	for i, scope := range config.Scopes {
		if scope == "offline_access" {
			// Offline_access is a special scope to request a refresh token that wont be there to display in the JIRA UI, so we skip it.
			continue
		}

		fmt.Printf("  %2d. %s\n", i+1, scope)
	}
```

internal/config/generator.go
```diff
@@ -93,6 +97,11 @@ type JiraCLIConfigGenerator struct {
 		mtls         struct {
 			caCert, clientCert, clientKey string
 		}
+		oauth struct {
+			accessToken  string
+			refreshToken string
+			cloudId      string
```
ccoVeille: You are using cloudID everywhere else


```suggestion
			cloudID      string
```

internal/config/generator.go
```diff
@@ -403,65 +498,75 @@ func (c *JiraCLIConfigGenerator) configureServerAndLoginDetails() error {
...
-			CaCert:     c.value.mtls.caCert,
-			ClientCert: c.value.mtls.clientCert,
-			ClientKey:  c.value.mtls.clientKey,
-		},
-	})
+	if c.jiraClient != nil {
+		config := c.generateJiraConfig()
+		c.jiraClient = api.Client(config)
+	}
 	info, err := c.jiraClient.ServerInfo()
```
ccoVeille: This part looks strange to me

So we had a non nil jiraClient, and we set a new one ?

This kind of code is usually when the jiraClient is nil

pkg/jira/cloud_id.go
```diff
@@ -0,0 +1,62 @@
...
+	AvatarURL string   `json:"avatarUrl"`
+}
+
+func (c *Client) GetCloudID() (string, error) {
+	envCloudID := os.Getenv("JIRA_CLI_CLOUD_ID")
+	if envCloudID != "" {
+		return envCloudID, nil
+	}
+
+	res, err := c.request(context.Background(), http.MethodGet, "https://api.atlassian.com/oauth/token/accessible-resources", nil, Header{
```
ccoVeille: You added a constant for this

christianarty:
Thanks @ccoVeille for the comments! Appreciate the extra look! I've addressed them all and tested it with the new revisions and things work as intended. @ankitpokhrel it now defaults to the keyring and fallsback to the filesystem storage. Please let me know if there is anything else we need to do in this PR otherwise i think its ready to 🚢

pkg/oauth/tokens.go
```diff
@@ -0,0 +1,167 @@
...
+	return &PersistentTokenSource{
+		clientID:        clientID,
+		clientSecret:    clientSecret,
+		storage:         keyringStorage,
+		fallbackStorage: fallbackFileSystemStorage,
+		usingFallback:   false,
+	}, nil
+}
+
+// Token implements oauth2.TokenSource interface.
```
ccoVeille:
```suggestion
// Token implements [oauth2.TokenSource] interface.
```

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,504 @@
...
+func printExpectedScopes(scopes []OAuthScope) {
+	var visibleScopes []OAuthScope
+	for _, scope := range scopes {
+		if scope.Visible {
+			visibleScopes = append(visibleScopes, scope)
+		}
+	}
+
+	// Sort by scope type (classic first, then granular) and then by name alphabetically
+	sort.Slice(visibleScopes, func(i, j int) bool {
```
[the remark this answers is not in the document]
christianarty: Oh, the more you know! Thanks for the info!

ccoVeille (approved):

christianarty:
@ankitpokhrel, whenever you get a chance could you drop a 🚢 . Would be greatly appreciated thanks!

ankitpokhrel:
> Hi @christianarty, first of all, amazing work and sorry for jumping in a bit late!
> 
> I ran some tests, and most things are working great. I noticed a few things tho:
> 
> 1. Listing epic items returns 401. Perhaps we are missing some permissions?
> 
> ```shell
> ➔ jira epic list ABC-3
> 
> jira: Received unexpected response '401 Unauthorized'.
> Please check the parameters you supplied and try again.
> ```
> 
> 2. The warning is shown every time when `oAuth` is not used. I think we only want to show warning if the type is `oAuth` and `api_server` is not set.
> 
> Image
> 3. Using `jira init` with `Authentication type: cloud` is failing with authentication err.
> 
> Image
> Could you please help check?
> 
> _(Feel free to ignore any copilot comment if that doesn't make sense)_
> 
> Thank you again for great work! 🎉


Hi @christianarty, while some issues appear to be resolved, issues 2 and 3 from the above list are still reproducible.

### Steps to reproduce:

- Check out the `main` branch, build the binary, and run `jira init` using `Authentication type: cloud` to complete the cloud configuration.
- Check out your branch, build the binary, and run any `jira` command.
    - You will see warnings related to 3LO, even though 3LO is not intended to be used at this stage.
- Run `jira init` again with `Authentication type: cloud`.
    - This fails with an error in your branch.

Basically, what we want to ensure is that the default behavior/backwards-compatibility remains unchanged.

api/client.go
```diff
@@ -6,40 +6,75 @@ import (
...
 	}
 
 	if config.Server == "" {
-		config.Server = viper.GetString("server")
+		apiServer := viper.GetString("api_server")
+		if apiServer != "" {
+			config.Server = apiServer
+		} else {
+			// Fallback to server URL if api_server is not set
+			cmdutil.Warn("api_server key is not set, falling back to server URL")
```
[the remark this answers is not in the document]
ankitpokhrel: I think copilot is correct in this case?

api/client.go
```diff
@@ -6,40 +6,75 @@ import (
 	"github.com/spf13/viper"
 	"github.com/zalando/go-keyring"
 
+	"github.com/ankitpokhrel/jira-cli/internal/cmdutil"
```
ankitpokhrel: Can we avoid using internal pkg here too?

pkg/oauth/tokens.go
```diff
@@ -0,0 +1,167 @@
+package oauth
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/ankitpokhrel/jira-cli/internal/cmdutil"
```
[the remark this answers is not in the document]
christianarty: Resolved in this commit specifically: https://github.com/ankitpokhrel/jira-cli/pull/877/commits/5828f1d2a24db42470ddd0938dab506f5de60536, I moved some stuff around to where I can still use the utils but avoid importing from internal

api/client.go
christianarty: https://github.com/ankitpokhrel/jira-cli/pull/877#discussion_r2679164647

christianarty:
> > Hi @christianarty, first of all, amazing work and sorry for jumping in a bit late!
> > I ran some tests, and most things are working great. I noticed a few things tho:
> > 
> > 1. Listing epic items returns 401. Perhaps we are missing some permissions?
> > 
> > ```shell
> > ➔ jira epic list ABC-3
> > 
> > jira: Received unexpected response '401 Unauthorized'.
> > Please check the parameters you supplied and try again.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. The warning is shown every time when `oAuth` is not used. I think we only want to show warning if the type is `oAuth` and `api_server` is not set.
> > 
> > Image
> > 3. Using `jira init` with `Authentication type: cloud` is failing with authentication err.
> > Image
> > Could you please help check?
> > _(Feel free to ignore any copilot comment if that doesn't make sense)_
> > Thank you again for great work! 🎉
> 
> Hi @christianarty, while some issues appear to be resolved, issues 2 and 3 from the above list are still reproducible.
> 
> ### Steps to reproduce:
> * Check out the `main` branch, build the binary, and run `jira init` using `Authentication type: cloud` to complete the cloud configuration.
> * Check out your branch, build the binary, and run any `jira` command.
>   
>   * You will see warnings related to 3LO, even though 3LO is not intended to be used at this stage.
> * Run `jira init` again with `Authentication type: cloud`.
>   
>   * This fails with an error in your branch.
> 
> Basically, what we want to ensure is that the default behavior/backwards-compatibility remains unchanged.

AH, thanks for that clarification, I did not do my original testing with a `config.yml` before these changes. Nice catch. I was able to replicate both of the issues you've described. 

1) The warning is shown every time when oAuth is not used
Resolved. Realized the logic was incorrect in adding in the `api_server` field. I moved it to check that only when we are doing oAuth stuff.
2) Using `jira init` with `Authentication type: cloud` is failing with authentication err
Resolved. The issue here was that previously when setting the cloud auth type, there was no distinction between oauth or cloud. It would just resolve to basic auth for cloud tokens. With the new change, that use case was not accounted for. The latest revision solves that issue
3) package files were importing from internal
Resolved. Moved some things around in [this commit specifically](https://github.com/ankitpokhrel/jira-cli/commit/5828f1d2a24db42470ddd0938dab506f5de60536), to allow for us to still use these utilities but not import them from the internal package. 

@ankitpokhrel Thanks for the comments, if there are no other issues, are we good to 🚢 ?

christianarty:
@ankitpokhrel anything else that needs to change in this PR?

vibbix:
Looking forward to this getting merged in! It's been a major blocker for me

christianarty:
@ankitpokhrel Any objections to merging this in?

PR #882, opened 2025-07-27
fix: csv format priority

ciricc:
Then the pipe used, code ignored the CSV format and made output just like the Plain enabled (because output is not TTY)
Example: 

```sh
jira issuse ls --csv >> output.csv
cat output.csv
```

Prints the plain in the old version instead of CSV format.


internal/view/issues_test.go
```diff
@@ -148,6 +148,51 @@ TEST-2	Story	Open	2020-12-13 14:05:20
...
+Story,TEST-2,This is another test,Open,,Person A,Normal,,2020-12-13 14:05:20,2020-12-13 14:07:20,"pat,mat"
+`
+	assert.Equal(t, expected, b.String())
+}
+
+func TestIssueRenderInCSVFormatWithoutHeaders(t *testing.T) {
+	var b bytes.Buffer
+
+	issue := IssueList{
+		Total:   2,
```
ankitpokhrel, line 178:
```suggestion

```

PR #886, opened 2025-08-07
feat: add unformatted flag to be able to avoid md parser

dydyamotya:
I encountered errors with markdown parsing when one uses:
1.` {*}text text{*} `to format bold text (added code style here to prevent rendering by markdown)
2. brackets inside brackets like so [[somestaff] Link text|example.com] 

I don't need fancy stuff, i just want to have info so I've added flag to bypass errors with parsing

I can't provide you real examples without breaking my NDA, so synthetic stuff.

How it should look like:
ответ нужно передавать в поле **INSURANCE_MODULE_RISK_GRADE**

How it looks like in raw:
ответ нужно передавать в поле `{*}INSURANCE_MODULE_RISK_GRADE{*}` (added code style here to prevent rendering by markdown)

How it looks like in plain or glamour:
²ÐµÑ<U+0082> Ð½Ñ<U+0083>Ð¶Ð½Ð¾ Ð¿ÐµÑ<U+0080>ÐµÐ´Ð°Ð²Ð°Ñ<U+0082>Ñ
<U+008C> Ð² Ð¿Ð¾Ð»Ðµ {}INSURANCE_MODULE_RISK_GRADE{}

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,11 +127,13 @@ Check `jira completion --help` for more info on setting up a bash/zsh shell comp
 
 #### Multiple projects
 
-The tool honors the `JIRA_CONFIG_FILE` environment variable, which specifies the location of the configuration file:
+You can load a specific configuration file by using the `--config/-c` flag, or by setting the `JIRA_CONFIG_FILE` environment variable to specify the file's location.
 
 ```sh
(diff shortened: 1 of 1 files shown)


dydyamotya:
Fixed formatting errors

PR #905, opened 2025-10-28
feat: add global non-interactive mode support

idleyoungman:
Add `--no-input` flag and `no_input` config setting to disable all interactive
prompts. When enabled, commands fail immediately with descriptive errors
instead of prompting for missing required fields, making the tool suitable
for programmatic usage, scripts, and CI/CD pipelines.

The feature can be enabled via:
- CLI flag: `--no-input`
- Config file: `no_input: true` in `~/.jira/.config.yml`
- Environment variable: `JIRA_NO_INPUT=true`

All commands that require user input now support non-interactive mode and
exit with status code 1 when required fields are missing.

### DIFF
--- a/internal/cmdutil/utils_test.go
+++ b/internal/cmdutil/utils_test.go
@@ -267,8 +267,6 @@ func TestGetSubtaskHandle(t *testing.T) {
 }
 
 func TestIsNoInputMode(t *testing.T) {
-	t.Parallel()
-
 	cases := []struct {
 		name     string
 		setup    func()
@@ -300,11 +298,7 @@ func TestIsNoInputMode(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		tc := tc
-
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
-
 			tc.setup()
 			assert.Equal(t, tc.expected, IsNoInputMode())
 		})
@@ -312,8 +306,6 @@ func TestIsNoInputMode(t *testing.T) {
 }
 
 func TestShouldPrompt(t *testing.T) {
-	t.Parallel()
-
 	cases := []struct {
 		name          string
 		localNoInput  bool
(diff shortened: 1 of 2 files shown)


internal/cmdutil/utils_test.go
```diff
@@ -264,3 +265,175 @@ func TestGetSubtaskHandle(t *testing.T) {
...
+			},
+			expected: false,
+		},
+		{
+			name: "prefers viper.Set over unset environment variable",
+			setup: func() {
+				viper.Reset()
+				viper.AutomaticEnv()
+				viper.SetEnvPrefix("jira")
+				os.Unsetenv("JIRA_NO_INPUT")
```
ccoVeille: This one could be removed 

```suggestion
```

I feel like you added it for avoiding possible issues with parallel testing

internal/cmd/issue/edit/edit.go
```diff
@@ -246,13 +253,27 @@ func handleUserAssign(project, key, assignee string, client *jira.Client) {
 	}
 }
 
+// hasNoChanges checks if any editable fields were provided in the params.
+func hasNoChanges(params *editParams) bool {
```
ccoVeille: Shouldn't it be this?


```suggestion
func hasNoChanges(params *editParams) bool {
```

And so, it could become a method


```suggestion
func (params *editParams) hasNoChanges() bool {
```

internal/cmdutil/utils_test.go
idleyoungman: Tried to address this in da24dda.

internal/cmd/issue/edit/edit.go
idleyoungman: Tried to address this in da24dda.

idleyoungman:
@ccoVeille Thank you for the careful review. Go is not my preferred language and it showed here! Appreciate the help. Pushed da24dda with your suggestions.

PR #919, opened 2025-12-08
feat(list): ✨  add max-results option and automatically page

wmcnamee-coreweave:
This change allows users to pass `--max-results` and multiple requests will be made, using the "nextPageToken" from each response to automatically iterate through pages. This provides an enhanced user experience over the need to manually provide `--paginate N:M` and make multiple CLI calls

Additionally, as part of this change, some of the search functions were refactored to use functional options, which DRY's the code a bit (there was duplication between v2 and v3 functions)


wmcnamee-coreweave:
@ankitpokhrel can you take a look at this PR please?

PR #923, opened 2025-12-10
Update golangci-lint to 2.6.x, and update config format

sarlalian:
Updated golangci-lint from 1.67 to 2.6, and update the config file to the new format.

Config file updated using `golangci-lint migrate` command

### DIFF
--- a/pkg/netrc/reader.go
+++ b/pkg/netrc/reader.go
@@ -30,7 +30,6 @@ func parseNetrc(data string) []netrcLine {
 	var nrc []netrcLine
 	var l netrcLine
 	inMacro := false
-lines:
 	for _, line := range strings.Split(data, "\n") {
 		if inMacro {
 			if line == "" {


pkg/netrc/reader.go
```diff
@@ -42,23 +43,23 @@ func parseNetrc(data string) []netrcLine {
...
 			// that matches […]. Once a match is made, the subsequent .netrc tokens
 			// are processed, stopping when the end of file is reached or another
-			// machine or a default token is encountered.”
+			// machine or a default token is encountered."
 			switch f[i] {
 			case "machine":
 				l = netrcLine{machine: f[i+1]}
 			case "default":
-				break
+				break lines
```
ankitpokhrel: [The spec](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html) says:

```
‘default’
This is the same as machine name except that default matches any name. 
There can be only one default token, and it must be after all machine tokens. 
```

In this context, the break statement is intended to exit only the inner loop so that we can evaluate whether the current token corresponds to the `default` case and then decide whether to continue iterating the outer loop. 

Can we just ignore the linting err  with `//nolint:staticcheck` directive for this case?

pkg/netrc/reader.go
```diff
@@ -30,6 +30,7 @@ func parseNetrc(data string) []netrcLine {
 	var nrc []netrcLine
 	var l netrcLine
 	inMacro := false
+lines:
```
ankitpokhrel:
```suggestion
```

pkg/netrc/reader.go
sarlalian: Nice catch, also, sorry for disappearing, ended up with a bunch of family things throughout the end of December / early January.

PR #925, opened 2025-12-18
fix: handle escaped commas in custom field array values

wbh1:
Custom field array values can now contain commas by escaping them with a backslash (\,). Previously, values like "WL: Tools, Development and Support" would be incorrectly split into multiple array elements.

For example, `jira issue create --custom work-category='WL: Tools, Development and Support' ...` would get split into `WL: Tools` and ` Development and Support`, so I'd get an error back from my Jira saying that those don't exist as valid options for that multi-select field.

Changes:
- Added splitUnescapedCommas() function to handle escaped commas
- Updated constructCustomFields() in create.go
- Updated constructCustomFieldsForEdit() in edit.go
- Added test coverage for both create and edit operations


geo-bl:
This bug is still on main and im also having this issue.

PR #971, opened 2026-03-25
fix: --plain flag should imply --table in sprint list

### LINKED ISSUE #968: sprint list: --plain flag should imply --table
## Description

When running `jira sprint list --state active --plain`, the `--plain` flag is ignored and the command enters TUI explorer mode instead of producing plain text output.

## Expected behavior

`--plain` should imply `--table` (i.e., bypass the TUI explorer and render a plain table), same as how it works when a `SPRINT_ID` is provided.

## Actual behavior

The `--plain` flag is only passed to `view.DisplayFormat.Plain` to control table decoration style, but does **not** participate in the TUI vs table mode decision. The branching logic in `sprintExplorerView()` only checks `--table`:

```go
// internal/cmd/sprint/list/list.go
if table || tui.IsDumbTerminal() || tui.IsNotTTY() {
    cmdutil.ExitIfError(v.RenderInTable())
} else {
    cmdutil.ExitIfError(v.Render())  // TUI explorer — always taken when --table is absent
}
```

So `--plain` without `--table` has no observable effect on the sprint list view.

## Suggested fix

Include `plain` in the condition:

```go
if table || plain || tui.IsDumbTerminal() || tui.IsNotTTY() {
```

This is consistent with user expectation: `--plain` means "give me plain text output", which inherently requires table mode rather than TUI.

## Workaround

Use `--table --plain` together:

```bash
jira sprint list --state active --table --plain --no-headers
```

YoanWai:
## Summary

- When using `jira sprint list --plain`, the `--plain` flag was ignored because only `--table` was checked in the TUI vs table mode decision
- This adds `plain` to the branching condition in `sprintExplorerView()` so that `--plain` bypasses the TUI explorer and renders a plain table

## Details

In `internal/cmd/sprint/list/list.go`, the condition that decides between TUI explorer and table output was:

```go
if table || tui.IsDumbTerminal() || tui.IsNotTTY() {
```

Changed to:

```go
if table || plain || tui.IsDumbTerminal() || tui.IsNotTTY() {
```

The `plain` variable was already read from flags earlier in the same function (line 215) and passed to `view.DisplayFormat`, so this is a minimal, safe addition.

Closes #968

### DIFF
--- a/internal/cmd/sprint/list/list.go
+++ b/internal/cmd/sprint/list/list.go
@@ -258,7 +258,7 @@ func sprintExplorerView(sprintQuery *query.Sprint, flags query.FlagParser, board
 	table, err := flags.GetBool("table")
 	cmdutil.ExitIfError(err)
 
-	if table || tui.IsDumbTerminal() || tui.IsNotTTY() {
+	if table || plain || tui.IsDumbTerminal() || tui.IsNotTTY() {
 		cmdutil.ExitIfError(v.RenderInTable())
 	} else {
 		cmdutil.ExitIfError(v.Render())


ankitpokhrel (approved):
Thank you!

PR #972, opened 2026-03-25
fix: --show-all-issues should preserve user JQL filter

### LINKED ISSUE #969: sprint list: --show-all-issues overwrites user-provided --jql/-q filter
## Description

When using `--show-all-issues` together with `-q` / `--jql`, the user-provided JQL filter is silently discarded.

For example:

```bash
jira sprint list --current --show-all-issues -q "assignee = currentUser()"
```

The `assignee` filter has no effect — all issues are returned regardless of assignee.

## Root cause

In `singleSprintView()` ([list.go](https://github.com/ankitpokhrel/jira-cli/blob/main/internal/cmd/sprint/list/list.go)):

```go
q, err := query.NewIssue(project, flags)   // q.Params().JQL = user's -q value
if sprintQuery.Params().ShowAllIssues {
    q.Params().JQL = "project IS NOT EMPTY"  // overwrites user's -q entirely
}
```

The same pattern appears in `getIssueQuery()` for the explorer view path.

`--show-all-issues` replaces the entire `JQL` field instead of only adjusting the project scope, so any user-supplied filter (assignee, status, labels via `-q`) is lost.

## Suggested fix

Prepend the project override while preserving the user's JQL:

```go
if sprintQuery.Params().ShowAllIssues {
    if q.Params().JQL != "" {
        q.Params().JQL = "project IS NOT EMPTY AND " + q.Params().JQL
    } else {
        q.Params().JQL = "project IS NOT EMPTY"
    }
}
```

This way `Raw()` in the JQL builder sees the `project` keyword, drops the default single-project filter, and the rest of the user's query is preserved.

## Workaround

Skip `--show-all-issues` and embed the project override directly in `-q`:

```bash
jira sprint list --current \
  -q "project IS NOT EMPTY AND assignee = currentUser()" \
  --order-by updated --table --plain
```

This works because `Raw()` already detects a `project` keyword in the JQL and removes the default project filter.

YoanWai:
## Summary
- When using `--show-all-issues` with `jira sprint list`, the flag now preserves any user-specified JQL filter
- Previously, `--show-all-issues` would overwrite the user's custom JQL with `project IS NOT EMPTY`, discarding any `--jql` argument
- The fix combines `project IS NOT EMPTY` with the user's JQL using `AND` when both are present

Closes #969

### DIFF
--- a/internal/cmd/sprint/list/list.go
+++ b/internal/cmd/sprint/list/list.go
@@ -102,7 +102,11 @@ func singleSprintView(sprintQuery *query.Sprint, flags query.FlagParser, boardID
 			return nil, err
 		}
 		if sprintQuery.Params().ShowAllIssues {
-			q.Params().JQL = "project IS NOT EMPTY"
+			if q.Params().JQL != "" {
+				q.Params().JQL = "project IS NOT EMPTY AND " + q.Params().JQL
+			} else {
+				q.Params().JQL = "project IS NOT EMPTY"
+			}
 		}
 		resp, err := client.SprintIssues(sprintID, q.Get(), q.Params().From, q.Params().Limit)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


ankitpokhrel:
Hi @YoanWai, thank you for working on this. This was fixed in this commit https://github.com/ankitpokhrel/jira-cli/pull/1014/changes/792a1bdf428c71e8859a09e4e509ff8f6536d826, so closing this PR. Thanks again!

PR #976, opened 2026-04-03
fix(custom-fields): preserve JSON objects/arrays in --custom values

Hernanduno:
Problem

Custom fields provided via --custom were always treated as plain strings in key paths that needed structured payloads. This broke User Picker fields (single and multi) and other custom fields requiring nested objects because Jira expected a JSON object/array, not an escaped string.

Motivation

Users editing issues with commands like the one below would hit Jira validation errors even when passing valid JSON.

Example command (values anonymized)

jira issue edit  DEMO-16216 --custom 'code-reviewer={"name":"redacted.user@example.com"}' --no-input

Before this fix (internal payload sent)

"customfield_xxxxx": [{"set": "{\"name\":\"redacted.user@example.com\"}"}]

After this fix (internal payload sent)

"customfield_xxxxx": [{"set": {"name": "redacted.user@example.com"}}]

What changed

- Added a dedicated parser in pkg/jira/customfield_parser.go that detects JSON containers (object/array) and unmarshals them while preserving existing behavior for non-container primitives/strings.

- Wired create flow to pass parsed object/array directly into fields payload (instead of forcing string).

- Wired edit flow to emit set operations with any typed value (object/array) so Jira receives correct structured JSON.

- Kept pkg/jira/customfield.go focused on type declarations; parser logic is now isolated in its own file.

Testing

- Added unit tests in pkg/jira/customfield_json_test.go for create and edit with deeply nested JSON structures.

- Verified with: go test ./pkg/jira

### DIFF
--- a/pkg/jira/create.go
+++ b/pkg/jira/create.go
@@ -240,6 +240,10 @@ func constructCustomFields(fields map[string]string, configuredFields []IssueTyp
 			if identifier != strings.ToLower(key) {
 				continue
 			}
+			if parsed, ok := parseCustomFieldJSONContainer(val); ok {
+				data.Fields.M.customFields[configured.Key] = parsed
+				continue
+			}
 
 			switch configured.Schema.DataType {
 			case customFieldFormatOption:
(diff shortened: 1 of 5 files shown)


caltuntas:
I had a chance to test this PR, it worked thanks , what is pending here to merge it into master? I see it complains about code formatting permissions, can someone support this ?

PR #980, opened 2026-04-10
fix: auto-detect non-TTY stdin to prevent segfault in edit, move, and delete

steven-range:
**What does this PR solve?**

Several commands panic with a segfault in the survey library when stdin is not a terminal (e.g. piped input, CI pipelines, scripted agents). The crash occurs because the survey library assumes interactive TTY input.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x2 pc=0x1034a8018]

goroutine 1 [running]:
github.com/AlecAivazis/survey/v2/terminal.(*Cursor).MoveNextLine(...)
```

`issue create` and `epic create` already handle this case by checking `cmdutil.StdinHasData()` and implicitly setting `noInput=true` when stdin is not a TTY. This PR applies the same pattern to three other commands:

- **`issue edit`**: implicitly sets `noInput=true` when stdin is not a TTY (matching `issue create` behavior)
- **`issue move`**: errors with a clear message when required args (issue key, desired state) are missing in non-TTY context
- **`issue delete`**: errors with a clear message when issue key arg is missing in non-TTY context

**How to test?**

Before this fix:
```sh
# Segfaults
echo "" | jira issue edit ISSUE-1 --custom story-point-estimate=1

# Segfaults (if args missing)
echo "" | jira issue move
echo "" | jira issue delete
```

After this fix:
```sh
# Works — implicitly uses --no-input behavior
echo "" | jira issue edit ISSUE-1 --custom story-point-estimate=1

# Also works — agents/CI calling without a TTY no longer need --no-input
jira issue edit ISSUE-1 --custom story-point-estimate=1  # in non-TTY context

# Clear error instead of segfault
echo "" | jira issue move
# Error: ISSUE-KEY argument is required in non-interactive mode

echo "" | jira issue delete
# Error: ISSUE-KEY argument is required in non-interactive mode
```

**Checklist**

- [x] I have added/updated enough tests related to my changes.
- [x] I have also manually checked and verified that my changes fix the issue and doesn't break any other functionalities.
- [x] My changes are backwards compatible.

Note: No test files exist for `internal/cmd/issue/edit/`, `move/`, or `delete/` (consistent with the rest of the cmd layer). Changes are minimal guards matching the existing pattern in `issue create`.

### DIFF
--- a/internal/cmd/issue/edit/edit.go
+++ b/internal/cmd/issue/edit/edit.go
@@ -60,6 +60,9 @@ func edit(cmd *cobra.Command, args []string) {
 	project := viper.GetString("project.key")
 
 	params := parseArgsAndFlags(cmd.Flags(), args, project)
+	if !params.noInput && cmdutil.StdinHasData() {
+		params.noInput = true
+	}
 	client := api.DefaultClient(params.debug)
 	ec := editCmd{
 		client: client,
--- a/internal/cmd/issue/delete/delete.go
+++ b/internal/cmd/issue/delete/delete.go
@@ -101,6 +101,10 @@ func (mc *deleteCmd) setIssueKey(project string) error {
 		return nil
 	}
 
+	if cmdutil.StdinHasData() {
+		return fmt.Errorf("ISSUE-KEY argument is required in non-interactive mode")
+	}
+
 	var ans string
 
 	qs := &survey.Question{
--- a/internal/cmd/issue/move/move.go
+++ b/internal/cmd/issue/move/move.go
@@ -180,6 +180,10 @@ func (mc *moveCmd) setIssueKey(project string) error {
 		return nil
 	}
 
+	if cmdutil.StdinHasData() {
+		return fmt.Errorf("ISSUE-KEY argument is required in non-interactive mode")
+	}
+
 	var ans string
 
 	qs := &survey.Question{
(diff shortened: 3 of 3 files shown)


internal/cmd/issue/edit/edit.go
```diff
@@ -60,6 +60,9 @@ func edit(cmd *cobra.Command, args []string) {
 	project := viper.GetString("project.key")
 
 	params := parseArgsAndFlags(cmd.Flags(), args, project)
+	if !params.noInput && cmdutil.StdinHasData() {
+		params.noInput = true
+	}
```
ccoVeille, line 65: Unless, I'm wrong it means this

```suggestion
	params.noInput = cmdutil.StdinHasData()
```

steven-range: I don't want to blindly overwrite.  If the user (or coding agent) sends `--no-input`, that should be retained regardless of `cmdutil.StdinHasData()`.  There may be an argument to simplify to what you've proposed, since piping something could indicate programmatic usage, but I don't think we can guarantee that.  Changing that would be more than a bug fix, which is the intended purpose here. 

It _could_ become `params.noInput = params.noInput || cmdutil.StdinHasData()`, but I think that obscures readability for no tangible benefit.

ccoVeille: Thanks for your detailed reply. You are right 😅👍

I missed the fact we don't want to reset noInput to false depending on the value of the stdin.

PR #1008, opened 2026-07-17
feat: add cf-access as supported auth type

josephvoss:
Adds a new auth type `cf-access` which allows using this tool with Cloudflare Access. Defines the new auth type, create a new RoundTripper to handle the login / token fetch flow, and set it as the client transport if enabled.

### Testing

End to end testing requires a Jira instance avaiable that's behind [Cloudflare Access]. We've tested the manually and confirmed the desired behavior, but I've also include a unit test to validate that the RoundTripper that sets the needed access header as well. I'm happy to include additional tests mocking the JWT issuance and parsing, but that should all be handled by the [cloudflared library].

### Implementation details

If this auth type is set, call the [cloudflared library] to make a HEAD request to the provided server URL with a `cloudflared/jira-cli` user agent. This returns a 302 to the access login URL, including the KeyID and AUD to fetch the needed JWT in the response output. We then call `token.FetchToken` to either retrieve the access token from the user's local cache if they've auth'd recently, or run the web-based login flow configured for the application policy. In either case an [Access JWT] is returned, which is passed with the [Cf-Access-Token] header by the new RoundTripper to authenticate to the backing Jira instance.

[Cloudflare Access]: https://developers.cloudflare.com/cloudflare-one/access-controls/policies/
[cloudflared library]: https://github.com/cloudflare/cloudflared
[access JWT]: https://developers.cloudflare.com/cloudflare-one/access-controls/applications/http-apps/authorization-cookie/application-token/
[Cf-Access-Token]: https://developers.cloudflare.com/cloudflare-one/tutorials/cli/#curl

---

Please let me know if you see any issues with this or have alternative ideas. This is a rebased change my team has been using for a couple years now, but it's at the point I'd like it to be upstreamed.


pkg/jira/client.go
```diff
@@ -11,9 +11,13 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/cloudflare/cloudflared/logger"
```
ankitpokhrel: Most of the indirect dependencies added in `go.sum` is because of this logger. Unless cloudflared logger is doing anything special for our usecase, can we perhaps directly utilize `zerolog` instead of using cloudflared logger? `zerolog` is already a transient dependency for cloudflared logger.

Something like this should work?

```diff
--- a/pkg/jira/client.go
+++ b/pkg/jira/client.go
@@ -18,2 +18,2 @@
-	"github.com/cloudflare/cloudflared/logger"
 	"github.com/cloudflare/cloudflared/token"
+	"github.com/rs/zerolog"

@@ -182,15 +182,17 @@
 	if c.AuthType != nil && *c.AuthType == AuthTypeCFAccess {
 		token.Init("jira-cli")
-		tempAccessURL, _ := url.Parse(c.Server)
-		appInfo, err := token.GetAppInfo(tempAccessURL)
 		cloudflaredLogger := zerolog.New(os.Stderr)
+		tempAccessURL, err := url.Parse(c.Server)
 		if err != nil {
+			log.Fatalf("failed to parse server URL: %v", err)
+		}
+		appInfo, err := token.GetAppInfo(tempAccessURL)
+		if err != nil {
+			log.Fatalf("failed to fetch access token: %v", err)
 		}
 		tok, err := token.FetchToken(tempAccessURL, appInfo, false, false, &cloudflaredLogger)
 		if err != nil {
+			log.Fatalf("failed to fetch access token: %v", err)
 		}
 		client.transport = NewCFAccessRoundTripper(tok, transport)
 	} else {
```

ankitpokhrel:
Hi @josephvoss, thanks for this! I don’t currently have the right setup to test this feature end-to-end, but the changes look good at a high level.

A few asks before we merge:

- `cloudflared/logger` appears to pull in quite a few unrelated indirect dependencies (gin, prometheus, websocket, etc). Could we see if there’s a way to avoid this so we don’t unnecessarily bloat go.sum?
- There’s a minor merge conflict due to some recent updates that needs to be resolved.
- Could you briefly document how to test this feature in the PR description? That would be helpful for future reference.

Thank you!

pkg/jira/client.go
josephvoss: Thanks for calling that out, I didn't realize we can just pass the raw zero logger here instead. Updated to match your suggested changes.

josephvoss:
Thank you @ankitpokhrel for giving this a look!

> * `cloudflared/logger` appears to pull in quite a few unrelated indirect dependencies (gin, prometheus, websocket, etc). Could we see if there’s a way to avoid this so we don’t unnecessarily bloat go.sum?

Great suggestion, addressed in https://github.com/ankitpokhrel/jira-cli/pull/1008#discussion_r3699655667

> * There’s a minor merge conflict due to some recent updates that needs to be resolved.

Addressed with the rebase now

> * Could you briefly document how to test this feature in the PR description? That would be helpful for future reference.

Updated the PR description to include how this works and links to the relevant documentation. I've also added a very minor unit test just to verify that the header is set correctly. Happy to remove if you think it's not valuable though.

I've also dropped the go version bump commit and the conditional header setting on the roundtripper. The full change can be seen with `git range-diff 1f3b2eff 831e7320 38d65c61 --creation-factor=100` if needed, but it's mostly pruning all the go.mod changes

ankitpokhrel (approved):
Thank you @josephvoss 🎉

PR #1018, opened 2026-08-17
fix(epic): migrate tree search to /search/jql (cloud 410 Gone)

hackerh3:
Jira Cloud removed the classic `/rest/api/2/search` endpoint (CHANGE-2046) — `jira epic tree` fails with 410 Gone on cloud. Migrate to the token-paginated `/rest/api/3/search/jql` API with automatic fallback to v2 `/search` for Server/DC (which has no v3 API).

Verified live against vectorgrp02.atlassian.net (XPE3-2444 renders) and covered by tests including token pagination and the DC fallback path.

https://claude.ai/code/session_012qTAdJSyqNg1aWb3PAF2A5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 on:
+  workflow_dispatch:
   pull_request:
     types: [opened, synchronize, reopened]
   release:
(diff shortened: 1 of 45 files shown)


hackerh3:
Apologies — opened against the wrong base by tooling; this targets code that only exists in my fork.

PR #1030, opened 2026-09-14
Fix deprecated postflight/verified cask stanzas

rethab:
Homebrew 6.0 deprecated the raw-Ruby `postflight` stanza (use `postflight_steps` instead) and the `verified:` url parameter (now a no-op — default URL verification covers it). Every `brew` command touching either generated cask (`jira-cli` and `jira-cli-snapshot`) prints warnings pointing users at the tap, but the tap files are generated here and get overwritten on the next release/merge, so they can't be fixed there.

GoReleaser's `homebrew_casks.hooks.post.install` always renders a raw `postflight do` block regardless of config (goreleaser/goreleaser#6870), and its fix (#6873) isn't merged yet. Emitting `postflight_steps` via `custom_block` is the documented workaround other projects have used for this. Note this does place the stanza at the top of the generated cask, which `brew style` flags as out-of-order (cosmetic only, doesn't affect install).

The hand-rolled `render-snapshot-cask.sh` template is updated the same way, plus dropping `verified:` from all four platform URLs.

Verified locally with `goreleaser check`, a `goreleaser release --snapshot` build, and `brew info` against the rendered casks in the actual tap — no more deprecation warnings.

https://claude.ai/code/session_014kcuW3F4bxv8UpVTi9GPPn


rethab:
Opened against the wrong repo by mistake (meant for my own fork's release tooling, not upstream). Sorry for the noise.

